PR #134, opened 2021-04-19
Dockerization

### LINKED ISSUE #137: Dockerization
**Describe the task**
For the AWS deployment plan, we need to dockerize the current app.

**Expected behavior**
1. Make the app dockerized
2. Update installation instructions

rathorevaibhav:
Fixes #137

## Summary

1. Dockerization is in place
2. Removed all major configs from zappa settings. All things moved to `.env`.
3. Entrypoint.sh to load data from Django fixture to create default tenant. Users won't have to set up default tenant anymore.
4. Updated installation instructions

## Test Plan

Nothing needed right now.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,5 @@
-#Django
-SECRET_KEY="" # generate from https://djecrety.ir/
+# django
+SECRET_KEY= # generate from https://djecrety.ir/
 DEBUG=False
 
 # database access credentials
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,31 @@
+#Django
+SECRET_KEY="my_secret_key"
+DEBUG=1
+ALLOWED_HOSTS=localhost 127.0.0.1 0.0.0.0 [::1] *
+
+# database access credentials
+DB_ENGINE=django_tenants.postgresql_backend
+APP_PORT=8001
+DB_HOST=db
+DB_NAME=plio
+DB_PORT=5432
+DB_USER=postgres
+DB_PASSWORD=
+
+# superuser details
+DJANGO_SU_NAME=vaibhav
+DJANGO_SU_EMAIL=vaibhav@coloredcow.com
+DJANGO_SU_PASSWORD=admin
+
+
+# plio cms token
+CMS_TOKEN=
+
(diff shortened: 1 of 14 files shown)


entrypoint.sh
```diff
@@ -1,11 +1,15 @@
...
 
-# start server
-python manage.py migrate && python manage.py runserver 0.0.0.0:${APP_PORT}
+# run migrations
+python manage.py migrate
 
+if [ ${APP_ENV} = "local" ]; then
+    python manage.py loaddata organizations/fixtures/default_tenant.yaml
+fi
 # create default tenant
```
dalmia, line 10: this documentation line should be before the code right?

entrypoint.sh
```diff
@@ -1,11 +1,15 @@
...
 set -e
 
-# start server
-python manage.py migrate && python manage.py runserver 0.0.0.0:${APP_PORT}
+# run migrations
+python manage.py migrate
 
+if [ ${APP_ENV} = "local" ]; then
+    python manage.py loaddata organizations/fixtures/default_tenant.yaml
+fi
```
dalmia, line 9: wouldn't we have to set up the public schema for staging and prod too?

.env.example
```diff
@@ -0,0 +1,27 @@
+# django
+APP_ENV='local' # local, staging, production
+SECRET_KEY= # generate from https://djecrety.ir/
+DEBUG=False
```
dalmia, line 4: by default should this be True?

rathorevaibhav: I think it depends on the user. having a different variable other than APP_ENV gives the flexibility to have any setting in place.

We can of course add the logic to make it default to True if app_env is local in settings.py. But I see that's more or less redundant.

entrypoint.sh
rathorevaibhav: yea we'll have to, I'm just thinking how that can happen. Given we only need public schema once and if we add a fixture with hardcoded domain, it only makes sense for Plio. Anyone else cloning this repo and setting up for themselves won't need those specific configs.

Let me work on this separately. #140

dalmia (approved):

PR #145, opened 2021-04-22
User access to org level plios

### LINKED ISSUE #144: User access to org level plios
**Describe the task**
A user who belongs to an org, should be able to view/play/edit/delete all the PUBLIC plios that are created by other users of the org.

A user who belongs to an org, should NOT be able to view/edit/delete the PRIVATE plios that are created by other users of the org.

A user who DOES NOT belong to an org, should not be able to access those org's plios - through frontend or through the API

A user, who is NOT the original creator of the plio, should be able to edit the plio, but the `created_by` field in that plio should still be equal to the original user who was the original creator of the plio

deepansh96:
Fixes #144

## Summary
- [x] A user who belongs to an org, should be able to view/play/edit/delete all the PUBLIC plios that are created by other users of the org.

- [x] A user who belongs to an org, should NOT be able to view/play/edit/delete the PRIVATE plios that are created by other users of the org.

- [x] A user who DOES NOT belong to an org, should not be able to access those org's plios - through frontend or through the API

- [x] A user, who is NOT the original creator of the plio, should be able to edit the plio, but the `created_by` field in that plio should still be equal to the original user who was the original creator of the plio
- [x] Allow std input and allow a docker container to be attached to the terminal - for debugging purposes

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,4 +20,4 @@ repos:
         # these are errors that will be ignored by flake8
         # check out their meaning here
         # https://flake8.pycqa.org/en/latest/user/error-codes.html
-        - "--ignore=E501,E203"
+        - "--ignore=E501,E203,W503"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,6 +15,8 @@ services: # the different images that will be running as containers
       - postgres_data:/var/lib/postgresql/data/
     restart: "on-failure"
   web: # service name
+    stdin_open: true
+    tty: true
     build: . # build the image for the web service from the dockerfile in parent directory.
     volumes:
       - .:/app # map data and files from parent directory in host to plio directory in docker container
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,6 +15,7 @@ services: # the different images that will be running as containers
       - postgres_data:/var/lib/postgresql/data/
     restart: "on-failure"
   web: # service name
+    # allows one to attach the terminal of a running container for debugging
     stdin_open: true
     tty: true
     build: . # build the image for the web service from the dockerfile in parent directory.
--- a/organizations/middleware.py
+++ b/organizations/middleware.py
@@ -8,13 +8,17 @@ class OrganizationTenantMiddleware(TenantMainMiddleware):
     Sets connection to either public or tenant schema based on `ORGANIZATION` HTTP header value.
     """
 
+    def get_organization_shortcode(self, request):
+        """
+        Returns the value of the `ORGANIZATION` HTTP header
+        """
+        return request.META.get("HTTP_ORGANIZATION", get_public_schema_name())
+
     def get_tenant(self, tenant_model, request):
         """
         Determines tenant by the value of the `ORGANIZATION` HTTP header.
         """
-        organization_shortcode = request.META.get(
-            "HTTP_ORGANIZATION", get_public_schema_name()
-        )
+        organization_shortcode = self.get_organization_shortcode(request)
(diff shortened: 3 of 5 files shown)


plio/views.py
```diff
@@ -94,6 +122,15 @@ def unique_users(self, request, uuid):
             ]
         )
 
+    def is_user_in_org(self, organization_shortcode):
+        org = Organization.objects.filter(shortcode=organization_shortcode)
+        if org:
+            org_id = org.first().id
+            return OrganizationUser.objects.filter(
+                organization=org_id, user=self.request.user.id
+            ).exists()
```
dalmia, line 131: why not directly do this check only?

plio/views.py
```diff
@@ -94,6 +122,15 @@ def unique_users(self, request, uuid):
             ]
         )
 
+    def is_user_in_org(self, organization_shortcode):
```
dalmia, line 125: this should not be inside plio/views.py - should be within some file in `users/`

.pre-commit-config.yaml
```diff
@@ -20,4 +20,4 @@ repos:
         # these are errors that will be ignored by flake8
         # check out their meaning here
         # https://flake8.pycqa.org/en/latest/user/error-codes.html
-        - "--ignore=E501,E203"
+        - "--ignore=E501,E203,W503"
```
dalmia, line 23: what is this error being ignored?

docker-compose.yml
```diff
@@ -15,6 +15,8 @@ services: # the different images that will be running as containers
       - postgres_data:/var/lib/postgresql/data/
     restart: "on-failure"
   web: # service name
+    stdin_open: true
+    tty: true
```
dalmia, line 19: one liners for what they do

organizations/middleware.py
```diff
@@ -8,13 +8,17 @@ class OrganizationTenantMiddleware(TenantMainMiddleware):
     Sets connection to either public or tenant schema based on `ORGANIZATION` HTTP header value.
     """
 
+    def get_organization_shortcode(self, request):
```
dalmia, line 11:
```suggestion
    def get_organization(self, request):
```

plio/views.py
```diff
@@ -46,17 +49,42 @@ class PlioViewSet(viewsets.ModelViewSet):
     lookup_field = "uuid"
 
     def get_queryset(self):
-        queryset = Plio.objects.filter(created_by=self.request.user)
-        return queryset
+        organization_shortcode = (
+            OrganizationTenantMiddleware.get_organization_shortcode(self, self.request)
```
dalmia, line 53:
```suggestion
            OrganizationTenantMiddleware.get_organization_shortcode(self.request)
```

dalmia (changes requested):
Broadly looks good to me - will test it now and add more feedback if any.

dalmia:
There is a bug - workspace switcher not working
I verified that the bug is due to this commit only by trying the workspace switcher in `shift-to-db` branch.

plio/views.py
```diff
@@ -46,17 +49,42 @@ class PlioViewSet(viewsets.ModelViewSet):
...
     def list_uuid(self, request):
         """Retrieves a list of UUIDs for all the plios"""
-        value_queryset = self.get_queryset().values_list("uuid", flat=True)
-        return Response(list(value_queryset))
+        queryset = self.get_queryset()
+        if not queryset or not queryset.exists():
+            return Response(
+                {"detail": "No plios found"},
+                status=status.HTTP_404_NOT_FOUND,
+            )
```
dalmia, line 86:
```suggestion
            return Response([])
```

dalmia (approved):

PR #155, opened 2021-05-07
FIX: handle empty org header and no org header, defaults to public

### LINKED ISSUE #154: Handle no org header or empty org header coming from the frontend and treat both cases as the `public` org in the backend

deepansh96:
Fixes #154


organizations/middleware.py
```diff
@@ -13,7 +13,11 @@ def get_organization(request):
         """
         Returns the value of the `ORGANIZATION` HTTP header
         """
-        return request.META.get("HTTP_ORGANIZATION", get_public_schema_name())
+        org = request.META.get("HTTP_ORGANIZATION", get_public_schema_name())
+        if org == "":
```
dalmia, line 17:
```suggestion
        if !org:
```

dalmia (commented):
Just one minor comment.

PR #162, opened 2021-05-12
Getting analytics access token from Auth0

rathorevaibhav:
Relates to https://github.com/avantifellows/plio-analytics/issues/4

## Summary

1. API endpoint for the frontend to make a call to retrieve analytics (cubejs) API token
2. Making API call to Auth0 with client id and the secret to retrieve a token

## Test Plan

Nothing right now.

## To do
1. Documentation update


users/views.py
```diff
@@ -202,3 +208,17 @@ def update_organization_user(sender, instance: OrganizationUser, **kwargs):
...
+    """ Makes a client credentials request to Auth0 app to get an access token. """
+    url = AUTH0_TOKEN_URL
+    payload = {
+        "grant_type": "client_credentials",
+        "client_id": AUTH0_CLIENT_ID,
+        "client_secret": AUTH0_CLIENT_SECRET,
+        "audience": AUTH0_AUDIENCE,
+    }
+    response = requests.post(url, data=payload)
+    return Response(response.json(), status=status.HTTP_200_OK)
```
dalmia, line 224: is the JWT being generated by Auth0 as well here?

rathorevaibhav: @dalmia yes, the JWT is being generated by Auth0. The response will be something like this:
```json
{
 "access_token": "token",
 "type": "Bearer",
 "expiry": 86400,
}
```

docs/ENV.md
```diff
@@ -99,3 +99,19 @@ Shortcode for the default tenant (e.g. plio)
 
 #### `DEFAULT_TENANT_DOMAIN`
 The domain for the default tenant (e.g. 0.0.0.0 locally, plio.in on production)
+
+
+### Auth0 for Plio Analytics
+While setting up Plio analytics, you need to make sure the following variables are also updated. These are responsible to fetch an access token from Auth0 Identity Provider.
```
dalmia, line 105: I am guessing we'll have to make another sub-domain for the production one right? 
Also, is any sort of payment set up for this?

users/views.py
```diff
@@ -202,3 +208,17 @@ def update_organization_user(sender, instance: OrganizationUser, **kwargs):
     async_to_sync(channel_layer.group_send)(
         user_group_name, {"type": "send_user", "data": user_data}
     )
+
+
+@api_view(["POST"])
+def retrieve_analytics_app_access_token(request):
+    """ Makes a client credentials request to Auth0 app to get an access token. """
```
dalmia, line 215:
```suggestion
    """Makes a client credentials request to Auth0 app to get an access token."""
```

users/views.py
dalmia, line 224: shouldn't we also check for error here?
And if it is successful, we don't need to pass the status right?

users/views.py
```diff
@@ -202,3 +208,17 @@ def update_organization_user(sender, instance: OrganizationUser, **kwargs):
...
+    url = AUTH0_TOKEN_URL
+    payload = {
+        "grant_type": "client_credentials",
+        "client_id": AUTH0_CLIENT_ID,
+        "client_secret": AUTH0_CLIENT_SECRET,
+        "audience": AUTH0_AUDIENCE,
+    }
+    response = requests.post(url, data=payload)
```
dalmia, line 223: any reason to not do?
```suggestion
    payload = {
        "grant_type": "client_credentials",
        "client_id": AUTH0_CLIENT_ID,
        "client_secret": AUTH0_CLIENT_SECRET,
        "audience": AUTH0_AUDIENCE,
    }
    response = requests.post(AUTH0_TOKEN_URL, data=payload)
```

dalmia (commented):
Left some comments.

users/views.py
rathorevaibhav: just a miss. I had it copy-pasted

users/views.py
rathorevaibhav: I think if the request fails, it will anyway result in an error. And that's how it should be. We can target handling the error message separately.

I believe we do need the 200 status as a parameter for returning response from rest framework.

PR #168, opened 2021-05-19
FIX: Backend will create a new retention string in case of a the first created session

### LINKED ISSUE #159: Retention Array is NULL in some cases

deepansh96:
…eated and sent to the client

Fixes #159


entries/serializers.py
```diff
@@ -65,6 +65,7 @@ def create(self, validated_data):
         session_answers = []
 
         # create the session answers
+        # create empty retention string if it doesn't exist
```
dalmia, line 68: the `if` below this is for when previous sessions exist - why should this be here?

PR #171, opened 2021-05-25
Api permissions scaffolding

### LINKED ISSUE #147: API permissions/access tasks
- prepare a list of APIs that only a super user should be able to access?
- A different user can still update the session and other data of another user right? We should fix that right?
- anyone can set any other user as superuser and access everything.

rathorevaibhav:
Fixes #147 

## Summary

Basic API permissions

## Test Plan

- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production

### DIFF
--- a/organizations/permissions.py
+++ b/organizations/permissions.py
@@ -7,4 +7,5 @@ class OrganizationPermission(permissions.BasePermission):
     """
 
     def has_permission(self, request, view):
+        """View-level permissions for organization. This determines whether the request can access organization instances or not."""
         return request.user.is_superuser
--- a/organizations/permissions.py
+++ b/organizations/permissions.py
@@ -7,4 +7,4 @@ class OrganizationPermission(permissions.BasePermission):
     """
 
     def has_permission(self, request, view):
-        return True
+        return request.user and request.user.is_superuser
--- a/organizations/permissions.py
+++ b/organizations/permissions.py
@@ -3,8 +3,8 @@ from rest_framework import permissions
 
 class OrganizationPermission(permissions.BasePermission):
     """
-    Global permission check for blocked IPs.
+    Global permission check for organizations.
     """
 
     def has_permission(self, request, view):
-        return request.user and request.user.is_superuser
+        return request.user.is_superuser
--- a/organizations/permissions.py
+++ b/organizations/permissions.py
@@ -3,7 +3,7 @@ from rest_framework import permissions
 
 class OrganizationPermission(permissions.BasePermission):
     """
-    Global permission check for organizations.
+    Permission check for organizations.
     """
 
     def has_permission(self, request, view):
--- a/organizations/permissions.py
+++ b/organizations/permissions.py
@@ -1,0 +1,10 @@
+from rest_framework import permissions
+
+
+class OrganizationPermission(permissions.BasePermission):
+    """
+    Global permission check for blocked IPs.
+    """
+
+    def has_permission(self, request, view):
+        return True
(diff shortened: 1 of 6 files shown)


deepansh96 (approved):

plio/permissions.py
```diff
@@ -0,0 +1,34 @@
...
+
+class PlioPermission(permissions.BasePermission):
+    """
+    Permission check for plios.
+    """
+
+    def has_permission(self, request, view):
+        return True
+
+    def has_object_permission(self, request, view, obj):
```
dalmia, line 15: similar to above, one-liners for what both functions are intended to do.

plio/permissions.py
```diff
@@ -0,0 +1,34 @@
...
+        if request.user.is_superuser:
+            return True
+
+        organization_shortcode = (
+            OrganizationTenantMiddleware.get_organization_shortcode(request)
+        )
+        if organization_shortcode == DEFAULT_TENANT_SHORTCODE:
```
dalmia, line 22: one line comment would be helpful - saying that when the plio is in the personal workspace, only the one who created it should be able to access.

plio/permissions.py
```diff
@@ -0,0 +1,34 @@
...
+            return request.user == obj.created_by
+
+        user_belongs_to_organization = OrganizationUser.objects.filter(
```
dalmia, line 25: similarly, one line comment here.

plio/permissions.py
dalmia, line 22: how does this work for the `play` endpoint? In that case, we should be granting read access, no?

users/permissions.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+
+    def has_permission(self, request, view):
+        if view.action in ["list", "create"]:
+            # only superuser can list or create users
+            return request.user.is_superuser
+
+        return True
+
+    def has_object_permission(self, request, view, obj):
```
dalmia, line 16: similar to above - one line comments for the functions.

organizations/permissions.py
```diff
@@ -0,0 +1,10 @@
+from rest_framework import permissions
+
+
+class OrganizationPermission(permissions.BasePermission):
+    """
+    Permission check for organizations.
+    """
+
+    def has_permission(self, request, view):
```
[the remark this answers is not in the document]
deepansh96: All of it is mentioned in the docs - https://www.django-rest-framework.org/api-guide/permissions/
I read and was able to understand the code
But yes a reference link should be mentioned as docstrings so readers know where to read and understand from

plio/permissions.py
dalmia: also, once it's done here, we won't need to do it in the `get_queryset` function right?

deepansh96: For the /play/ endpoint, the `has_object_permission` is not called. That method is not called for list views. This is only called when a put/post/patch request is called.
In the /play/ endpoint, we're not doing any putting/posting/patching. we're only listing. So for that case it skips this `has_object_permission`
But if someone is editing a plio, in that case this method is called

deepansh96: And no, we'll have to do it in get_queryset function as well.
Because in get_queryset, we need to filter plios in a different way depending on the organization shortcode and other conditions

plio/permissions.py
```diff
@@ -10,18 +10,24 @@ class PlioPermission(permissions.BasePermission):
...
+
         if request.user.is_superuser:
             return True
 
         organization_shortcode = (
             OrganizationTenantMiddleware.get_organization_shortcode(request)
         )
         if organization_shortcode == DEFAULT_TENANT_SHORTCODE:
+            # This is user's personal workspace in the shared DB schema.
+            # Only plios created by the user are allowed to access.
```
dalmia, line 27:
```suggestion
            # this is the user's personal workspace in the public DB schema
            # only plios created by the user can be accessed
```

PR #172, opened 2021-05-26
Superuser fix and organization-user role check

rathorevaibhav:
## Summary

1. Superuser and staff fields are readonly.
2. Checks for organization user

## Test Plan
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,3 +26,6 @@ Visit the [REST API guide](docs/REST-API.md) to connect your Plio's frontend set
 - [Zappa Settings](docs/ZAPPA-SETTINGS.md)
 - [Soft Delete](docs/SOFT-DELETE.md)
 - [Google OAuth2](docs/oauth/GOOGLE-OAUTH2.md)
+
+## Product Guide
+To know more about features and updates on Plio, visit the [product guide on our YouTube channel](https://www.youtube.com/playlist?list=PL3U0Jqw-piJgw2hSpuAZym4K1_Tb0RTRV).
--- a/users/models.py
+++ b/users/models.py
@@ -90,6 +90,15 @@ class User(SafeDeleteModel, AbstractUser):
     def __str__(self):
         return "%d: %s" % (self.id, self.name)
 
+    def get_role_for_organization(self, organization_id):
+        organization_user = OrganizationUser.objects.filter(
+            organization_id=organization_id, user_id=self.id
+        ).first()
+        if not organization_user:
+            return None
+
+        return organization_user.role
+
 
 class UserMeta(SafeDeleteModel):
     _safedelete_policy = SOFT_DELETE
--- a/users/permissions.py
+++ b/users/permissions.py
@@ -1,4 +1,5 @@
 from rest_framework import permissions
+from users.models import OrganizationUser
 
 
 class UserPermission(permissions.BasePermission):
@@ -17,3 +18,38 @@ class UserPermission(permissions.BasePermission):
     def has_object_permission(self, request, view, obj):
         """Object-level permissions for user. This determines whether the request can access a user instance or not."""
         return request.user.is_superuser or request.user == obj
+
+
+class OrganizationUserPermission(permissions.BasePermission):
+    """
+    Permission check for organization-user mapping.
+    """
+
+    def has_permission(self, request, view):
+        """View-level permissions for organization-user. This determines whether the request can access organization-user instances or not."""
+        if request.user.is_superuser:
+            return True
+
+        if view.action == "create":
+            # only organization admins can add a user to their organization
+            user_is_organization_admin = OrganizationUser.objects.filter(
+                organization=request.data["organization"],
+                user=request.user.id,
+                role__name="org-admin",
+            ).exists()
+            return user_is_organization_admin
+
+        return True
+
+    def has_object_permission(self, request, view, obj):
+        """Object-level permissions for organization-user. This determines whether the request can access an organization-user instance or not."""
+        if request.user.is_superuser:
+            return True
+
+        # only organization admins are allowed to access organization-user instance
+        user_is_organization_admin = OrganizationUser.objects.filter(
+            organization=request.data["organization"],
+            user=request.user.id,
+            role__name="org-admin",
+        ).exists()
+        return user_is_organization_admin
--- a/users/permissions.py
+++ b/users/permissions.py
@@ -1,5 +1,5 @@
 from rest_framework import permissions
-from users.models import OrganizationUser, Role
+from users.models import Role
 
 
 class UserPermission(permissions.BasePermission):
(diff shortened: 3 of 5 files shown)


users/permissions.py
```diff
@@ -17,3 +18,38 @@ def has_permission(self, request, view):
...
+        """View-level permissions for organization-user. This determines whether the request can access organization-user instances or not."""
+        if request.user.is_superuser:
+            return True
+
+        if view.action == "create":
+            # only organization admins can add a user to their organization
+            user_is_organization_admin = OrganizationUser.objects.filter(
+                organization=request.data["organization"],
+                user=request.user.id,
+                role__name="org-admin",
```
dalmia, line 38: there's a `super-admin` role too, right?

users/permissions.py
```diff
@@ -17,3 +18,38 @@ def has_permission(self, request, view):
...
+    """
+    Permission check for organization-user mapping.
+    """
+
+    def has_permission(self, request, view):
+        """View-level permissions for organization-user. This determines whether the request can access organization-user instances or not."""
+        if request.user.is_superuser:
+            return True
+
+        if view.action == "create":
```
dalmia, line 33: why only create? What about update/delete?

users/permissions.py
```diff
@@ -17,3 +18,38 @@ def has_permission(self, request, view):
...
+    def has_object_permission(self, request, view, obj):
+        """Object-level permissions for organization-user. This determines whether the request can access an organization-user instance or not."""
+        if request.user.is_superuser:
+            return True
+
+        # only organization admins are allowed to access organization-user instance
+        user_is_organization_admin = OrganizationUser.objects.filter(
+            organization=request.data["organization"],
+            user=request.user.id,
+            role__name="org-admin",
```
dalmia, line 53: same point about `super-admin` as above

users/permissions.py
dalmia: cannot update things if I am `super-admin` of an org right now.

rathorevaibhav: I think there is some confusion here also.

So what's the difference between `super-admin` and `org-admin` of an organization?

Is `super-admin` supposed to be same as `superuser`?

users/permissions.py
rathorevaibhav: Makes sense.

README.md
```diff
@@ -26,3 +26,6 @@ Visit the [REST API guide](docs/REST-API.md) to connect your Plio's frontend set
 - [Zappa Settings](docs/ZAPPA-SETTINGS.md)
 - [Soft Delete](docs/SOFT-DELETE.md)
 - [Google OAuth2](docs/oauth/GOOGLE-OAUTH2.md)
+
+## Product Guide
+To know more about features and updates on Plio, visit the [product guide on our YouTube channel](https://www.youtube.com/playlist?list=PL3U0Jqw-piJgw2hSpuAZym4K1_Tb0RTRV).
```
dalmia, line 31:
```suggestion
To know more about the features and updates on Plio, visit the [product guide on our YouTube channel](https://www.youtube.com/playlist?list=PL3U0Jqw-piJgw2hSpuAZym4K1_Tb0RTRV).
```

users/models.py
```diff
@@ -90,6 +90,15 @@ def name(self):
     def __str__(self):
         return "%d: %s" % (self.id, self.name)
 
+    def get_role_for_organization(self, organization_id):
+        organization_user = OrganizationUser.objects.filter(
```
dalmia, line 94:
```suggestion
        """Returns the user's role within the organization provided. Returns None if the user is not part of the organization"""
        organization_user = OrganizationUser.objects.filter(
```

users/permissions.py
```diff
@@ -30,26 +30,48 @@ def has_permission(self, request, view):
...
-        return True
+        user_organization_role = request.user.get_role_for_organization(
+            request.data["organization"]
+        )
+        if user_organization_role.name not in ["org-admin", "super-admin"]:
+            # user doesn't belong to the queried organization
+            # or doesn't have sufficient role within organization
+            return False
+
+        requested_role = Role.objects.filter(id=request.data["role"]).first()
```
dalmia, line 46: what is request.data doesn't have any role value?

dalmia: what if the provided role Id is wrong?

users/permissions.py
```diff
@@ -30,26 +30,48 @@ def has_permission(self, request, view):
...
-        if view.action == "create":
-            # only organization admins can add a user to their organization
-            user_is_organization_admin = OrganizationUser.objects.filter(
-                organization=request.data["organization"],
-                user=request.user.id,
-                role__name="org-admin",
-            ).exists()
-            return user_is_organization_admin
+        # when listing, a user should only see organization-user mapping for the organizations they have access to.
+        # this is handled by get_queryset
```
dalmia, line 34: can you please point me to where it is being handled inside get_queryset? I am not able to find anything relevant.

users/permissions.py
```diff
@@ -30,26 +30,48 @@ def has_permission(self, request, view):
...
+
+        # only super-admin and org-admin can access organization_user instance
+        if user_organization_role.name not in ["super-admin", "org-admin"]:
+            return False
+
+        # super-admins can add users with org-admin and org-view roles to their organization
+        if user_organization_role.name == "super-admin":
+            return obj.role.name in ["org-admin", "org-view"]
+
+        # or-admins can add users with org-view role to their organization
```
dalmia, line 75:
```suggestion
        # org-admins can add users with org-view role to their organization
```

users/permissions.py
rathorevaibhav: I thought about it but felt like a role must be a required value in this query. I'm not sure in which case it won't be provided.

I agree on incorrect role id. That check needs to be in place. I'll see what's the right fix.

users/permissions.py
rathorevaibhav: This hasn't been added yet. working on this.

users/permissions.py
```diff
@@ -60,9 +67,15 @@ def has_object_permission(self, request, view, obj):
...
-        user_organization_role = request.user.get_role_for_organization(
-            request.data["organization"]
-        )
+        if view.action == "retrieve":
+            return True
+
+        if view.action == "destroy":
+            organization_id = obj.organization_id
+        else:
+            organization_id = request.data["organization"]
```
dalmia, line 76: we return true above already if it is any of the two actions right?

users/views.py
```diff
@@ -97,6 +97,21 @@ class OrganizationUserViewSet(viewsets.ModelViewSet):
     queryset = OrganizationUser.objects.all()
```
dalmia, line 97: don't we have to delete this?

users/permissions.py
```diff
@@ -17,3 +18,73 @@ def has_permission(self, request, view):
...
+
+        user_organization_role = request.user.get_role_for_organization(organization_id)
+
+        # only super-admin and org-admin can access organization_user instance
+        if user_organization_role.name not in ["super-admin", "org-admin"]:
+            return False
```
dalmia, line 82: this check is no longer needed right?

PR #179, opened 2021-06-05
Support for subjective questions

### LINKED ISSUE #178: Support for subjective questions

dalmia:
Fixes #178 

## Summary
- updates the list of valid question types
- adds 2 new fields to the question table:
    - `has_char_limit`: whether the question has any character limit
    - `max_char_limit`: what is the maximum character limit if it has one

TODO:
- [x] test the dashboard front (downloading reports etc.)

## Test Plan

- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production


plio/models.py
```diff
@@ -95,6 +95,8 @@ class Question(SafeDeleteModel):
     correct_answer = models.TextField(null=True)
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
+    has_char_limit = models.BooleanField(default=False)
+    max_char_limit = models.FloatField(null=True)
```
deepansh96, line 99: can't we just use `max_char_limit`?
As it's null by default, we can assume that this question doesn't have any `max_char_limit` set
That way we won't need another field in our model

PR #181, opened 2021-06-08
Feature/testing

### LINKED ISSUE #180: Set up automated testing
**Describe the task**
As the Plio system is getting stable, we need to start writing test cases to test code stability at all times and avoid unprecedented issues.

**Expected behavior**
Set up Django unit test cases.

**References**
https://docs.djangoproject.com/en/3.2/intro/tutorial05/

rathorevaibhav:
Fixes #180 

## Summary

1. SMS_DRIVER introduced to avoid SMS sending when testing.
2. Pre-commit CI yml merged with test cases CI. Instead of two CI runs, we'll have two jobs within one CI. More or less same.
3. Removed `TenantMainMiddleware` from middleware list as it was the reason for 404 errors in all test cases. Since, Organization middleware inherits the TenantMainMiddleware, it is placed on top now.
4. Test case added for User Authentication

## Test Plan

- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production

## Additional
Because of middleware changes, I'll need help in testing functionalities manually. I've tested locally and there doesn't seem to be any issue.


.env.example
```diff
@@ -59,3 +59,8 @@ EMAIL_HOST_USER= # sender's email-id
 # passwords. More on that here:
 # https://dev.to/abderrahmanemustapha/how-to-send-email-with-django-and-gmail-in-production-the-right-way-24ab
 EMAIL_HOST_PASSWORD=
+
+# The driver for sending SMSs. Possible values are `sns` or `log`.
+# Use `sns` to have AWS SNS support. The AWS credentials must be present for this.
+# Use `log` to log SMSs into a file instead. Recommended for development mode.
+SMS_DRIVER='sns'
```
dalmia, line 66: you'll need to update .ENV.md - that's where we have been keeping the main details for each env variable

.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
...
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+
+      - name: Run Test Cases
+        env:
+          DB_HOST: 127.0.0.1
+          DB_NAME: github_actions
+          DB_USER: postgres
+          DB_PASSWORD: postgres
+          SECRET_KEY: TESTING_KEY
```
dalmia, line 52: is `TESTING_KEY` a valid value or a placeholder for something else?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
...
+        # command to run tests and generate coverage metrics
+        run: coverage run manage.py test
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
dalmia, line 59: need to document somewhere this new variable no?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
...
+      # Install Python. This matches the Python version in Dockerfile.
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Install Dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
```
dalmia, line 44: can't we directly install using the Dockerfile on the new instance?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
...
+    name: Test cases
+    runs-on: ubuntu-latest
+    services:
+      # Creates a postgres docker where migrations will run.
+      db:
+        image: postgres:12.3-alpine
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: github_actions
```
dalmia, line 27: wouldn't a better name would be `github_actions_testing`

users/tests.py
```diff
@@ -1 +1,48 @@
-# Create your tests here.
+from rest_framework.test import APIClient
+from rest_framework.test import APITestCase
+from users.models import OneTimePassword, User
+from plio.settings import API_APPLICATION_NAME
+from oauth2_provider.models import Application
+from django.core.management import call_command
+from django.urls import reverse
+from rest_framework import status
```
dalmia, line 8:
```suggestion
from rest_framework.test import APIClient
from rest_framework.test import APITestCase
from rest_framework import status
from oauth2_provider.models import Application
from django.core.management import call_command
from django.urls import reverse
from users.models import OneTimePassword, User
from plio.settings import API_APPLICATION_NAME
```

users/tests.py
```diff
@@ -1 +1,48 @@
...
+
+
+class BaseTestCase(APITestCase):
```
dalmia, line 11: one line description

users/tests.py
```diff
@@ -1 +1,48 @@
...
+    def setUp(self):
+        self.client = APIClient()
+        call_command("pliosuperuser")
+        self.super_user = User.objects.first()
```
dalmia, line 15: is this accurate? what if the first object is not a super user? or since we always create a superuser account upon installing first, this is okay?

users/tests.py
```diff
@@ -1 +1,48 @@
...
+        Application.objects.create(
+            name=API_APPLICATION_NAME,
+            redirect_uris="",
+            user=self.super_user,
+            client_type=Application.CLIENT_CONFIDENTIAL,
+            authorization_grant_type=Application.GRANT_AUTHORIZATION_CODE,
+        )
+
+
+class UserAuthenticationTestCase(BaseTestCase):
```
dalmia, line 25: this tests only the OTP part right? Not the G-auth one? Would it be better to clarify that in the name?

dalmia: + one line description

users/tests.py
```diff
@@ -1 +1,48 @@
...
+from oauth2_provider.models import Application
+from django.core.management import call_command
+from django.urls import reverse
+from rest_framework import status
+
+
+class BaseTestCase(APITestCase):
+    def setUp(self):
+        self.client = APIClient()
+        call_command("pliosuperuser")
```
dalmia, line 14: is the `pliosuperuser` command declared somewhere?

users/tests.py
```diff
@@ -1 +1,48 @@
...
+        self.super_user = User.objects.first()
+        Application.objects.create(
+            name=API_APPLICATION_NAME,
+            redirect_uris="",
+            user=self.super_user,
+            client_type=Application.CLIENT_CONFIDENTIAL,
+            authorization_grant_type=Application.GRANT_AUTHORIZATION_CODE,
```
dalmia, line 21: can you give some reference for this bit?

users/tests.py
```diff
@@ -1 +1,48 @@
...
+        )
+
+
+class UserAuthenticationTestCase(BaseTestCase):
+    def setUp(self):
+        super().setUp()
+        self.user_mobile = "+919876543210"
```
dalmia, line 28: shouldn't we make this some env variable? 
Also, if we detach SNS from this using SMS_DRIVER - we won't be able to guarantee that it works right? is this how people usually test an SMS functionality?

users/tests.py
```diff
@@ -1 +1,48 @@
...
+
+    def test_guest_can_request_for_otp(self):
+        response = self.client.post(
+            reverse("request_otp"), {"mobile": self.user_mobile}
+        )
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        otp_exists = OneTimePassword.objects.filter(mobile=self.user_mobile).exists()
+        self.assertTrue(otp_exists)
```
dalmia, line 37: shouldn't the test actually be:
- get the OTP received in the response
- test that the received OTP is present and is associated with the user mobile number?

users/tests.py
```diff
@@ -1 +1,48 @@
...
+    def test_guest_can_verify_otp(self):
+        # request otp
+        self.client.post(reverse("request_otp"), {"mobile": self.user_mobile})
+
+        # fetch otp
+        otp = OneTimePassword.objects.filter(mobile=self.user_mobile).first()
+        response = self.client.post(
+            reverse("verify_otp"), {"mobile": self.user_mobile, "otp": otp.otp}
+        )
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
dalmia, line 48: What is the request status when the OTP does not match?

We should also include a check when the wrong OTP is sent right? to verify that it returns an error?

rathorevaibhav: True. That's right suggestion. I'll add.

.github/workflows/ci.yml
rathorevaibhav: yeah it's a valid value, but i've added a random string now. We don't need to use an env variable I believe as it's only for testing purpose.

.github/workflows/ci.yml
rathorevaibhav: Since it's only required for maintaining coverage report, I don't think it fits well in ENV.md.

I've added some comments above this line. I think that should be sufficient.

users/tests.py
rathorevaibhav: Yeah, this is just a unit test case that assumes the third-party service works fine. We will have different test cases to test APIs.

It's a random mobile number (9 to 0 in decreasing order). Only for testing. If needed, we can have it as a variable to reuse but I don't think it fits as env variable.

users/tests.py
rathorevaibhav: We don't receive OTP in API response. It goes directly to the user. I believe that's how it should be.

Request OTP endpoint returns everything except the OTP text.

.github/workflows/ci.yml
rathorevaibhav: I dont think we can run tests within a container. That's what we are doing at GitHub level in our other deploy_* github actions.

We will need a running instance. Either this or use docker-compose at Github level. I think we can target it separately.
Here it is: #182

.github/workflows/ci.yml
```diff
@@ -0,0 +1,62 @@
+name: Plio CI
+
+on:
+  pull_request:
+  push:
+    branches: [master]
```
dalmia, line 6: won't this be ["master"]

.github/workflows/ci.yml
```diff
@@ -46,15 +46,17 @@ jobs:
...
-          SECRET_KEY: TESTING_KEY
+          SECRET_KEY: wpurj&oym6m@kcp(m&z(q-g0bo-r*+!f_&j(94di8j&_j4m%2s  # random secret key
         # command to run tests and generate coverage metrics
         run: coverage run manage.py test
 
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
         with:
+          # The codecov token to upload report. Can be retrieved from your Codecov dashboard.
+          # This should be set from `GitHub Repo > Settings > Environment Variables`.
```
dalmia, line 60:
```suggestion
          # This should be set from `GitHub Repo > Settings > Secrets`.
```

users/tests.py
```diff
@@ -1,28 +1,31 @@
...
 from oauth2_provider.models import Application
-from django.core.management import call_command
 from django.urls import reverse
-from rest_framework import status
+from users.models import OneTimePassword
+from plio.settings import API_APPLICATION_NAME
 
 
 class BaseTestCase(APITestCase):
+    """Base class that set up generic pre-requisites for all further test classes"""
```
dalmia, line 11:
```suggestion
    """Base class that sets up generic pre-requisites for all further test classes"""
```

.github/workflows/ci.yml
rathorevaibhav: It works the same in yml format.

PR #187, opened 2021-06-13
NEW: image model + upload to s3

### LINKED ISSUE #186: Support for adding images to questions

deepansh96:
Fixes #186
Depends on [this](https://github.com/avantifellows/plio-frontend/pull/331) PR
## Test Plan

- [x] Tested locally
- [x] Tested on staging
- [ ] Tested on production


docs/ENV.md
```diff
@@ -70,6 +70,9 @@ AWS secret access key.
 #### `AWS_REGION`
 Region of the AWS IAM user.
 
+#### `AWS_STORAGE_BUCKET_NAME`
```
dalmia, line 73: have these been made different for staging and prod?

docs/ENV.md
```diff
@@ -70,6 +70,9 @@ AWS secret access key.
...
+AWS bucket where Django storages uploads the files
```
dalmia, line 74: seems like some typo here.

plio/models.py
```diff
@@ -6,6 +6,20 @@
 from plio.config import plio_status_choices, item_type_choices, question_type_choices
 
 
+class Image(SafeDeleteModel):
+    _safedelte_policy = SOFT_DELETE
+
+    image_url = models.ImageField("Image", upload_to="images")
```
dalmia, line 12:
```suggestion
    url = models.ImageField("Image", upload_to="images")
```

dalmia: I think we also need to add alt text - all platforms have it and it's an important element of accessibility.

dalmia: we can set some default alt text and might not need to provide option to modify it - but at least something should be there.

plio/models.py
```diff
@@ -6,6 +6,20 @@
...
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        db_table = "image"
+
+    def __str__(self):
+        return f"Image id {self.id}"
```
dalmia, line 20: consistency with others
```suggestion
        return f"{self.id}: {self.url}"
```

plio/serializers.py
```diff
@@ -1,9 +1,20 @@
...
 from users.models import User
 from users.serializers import UserSerializer
 
 
+class ImageSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Image
+        fields = [
+            "id",
+            "image_url",
```
dalmia, line 12:
```suggestion
            "url",
```

plio/settings.py
```diff
@@ -309,3 +309,9 @@
 EMAIL_HOST_PASSWORD = os.environ.get("EMAIL_HOST_PASSWORD")
 
 SMS_DRIVER = os.environ.get("SMS_DRIVER")
+
+# file storage for uploaded images
+DEFAULT_FILE_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
+AWS_STORAGE_BUCKET_NAME = os.environ.get("AWS_STORAGE_BUCKET_NAME")
+AWS_QUERYSTRING_AUTH = False
```
dalmia, line 316: Screenshot 2021-06-13 at 9 38 54 PM
false is for only public s3 buckets right? We shouldn't have it public right?

plio/views.py
```diff
@@ -346,3 +347,24 @@ def duplicate(self, request, pk):
...
+    destroy: Soft delete a image file entry, along
+             with the uploaded S3 file
+    """
+
+    queryset = Image.objects.all()
+    serializer_class = ImageSerializer
+
+    def perform_destroy(self, instance):
+        instance.image_url.delete()
+        instance.delete()
```
dalmia, line 370: why is this needed if we didn't do this for VideoViewSet?

plio/models.py
deepansh96: What is alt text?

docs/ENV.md
deepansh96: Storages?
No. Django storages is the package name which handles the s3 storage

docs/ENV.md
deepansh96: no not yet, will do

plio/settings.py
deepansh96: If we don't have a public access enabled for the images, the images will not load on the client, or any browser even. Earlier the bucket was not public, it had no policy for any public user to read the files. And testing during that, I was not able to load the files on the plio

plio/views.py
deepansh96: Added `instance.image_url.delete()` to delete the image file from s3 as well. Without it, only the entry in the table is soft deleted.

docs/ENV.md
deepansh96: done

docs/ENV.md
dalmia: then better to use `django-storages` or else it's weird to read.

plio/models.py
dalmia: https://www.w3schools.com/tags/att_img_alt.asp

plio/settings.py
dalmia: bucket level private, file level public

plio/views.py
dalmia: create unique name for new image file

dalmia: revert not changing from s3 and update comments on frontend about deleting from s3 too

plio/views.py
```diff
@@ -366,5 +366,4 @@ class ImageViewSet(viewsets.ModelViewSet):
     serializer_class = ImageSerializer
 
     def perform_destroy(self, instance):
-        instance.image_url.delete()
         instance.delete()
```
dalmia: this whole thing is no longer needed right?

plio/models.py
```diff
@@ -2,10 +2,45 @@
 from django.db import models
 import string
 import random
+import os
 from safedelete.models import SafeDeleteModel, SOFT_DELETE
 from plio.config import plio_status_choices, item_type_choices, question_type_choices
 
 
+class Image(SafeDeleteModel):
```
dalmia, line 10: the only thing missing is tests :P

plio/permissions.py
```diff
@@ -14,7 +15,14 @@ def has_permission(self, request, view):
...
     def has_object_permission(self, request, view, obj):
-        """Object-level permissions for plio. This determines whether the request can access a plio instance or not."""
+        """Object-level permissions for plio/item/question. This determines whether the request can access a plio/item/question instance or not."""
+
+        # if the requested object is an Item or Question, filter the Plio record that it belongs to
+        # and use the permission logic on that Plio instance
+        if isinstance(obj, Question):
+            obj = Plio.objects.filter(id=obj.item.plio.id).first()
+        elif isinstance(obj, Item):
+            obj = Plio.objects.filter(id=obj.plio.id).first()
```
dalmia, line 25: was this added just for sanity or did something broke that led you to add this?
the permission class has only been added to `QuestionViewSet` and not `ItemViewSet`-  any reason?
Any reason for not adding this for ImageViewSet too?

plio/tests.py
```diff
@@ -135,3 +135,106 @@ def setUp(self):
...
+
+class ImageTestCase(PlioTestCase):
+    """Tests the Image CRUD functionality."""
+
+    def setUp(self):
+        # seed some plios
+        super().setUp()
+
+        # seed some items and questions
+        self.plio_1 = Plio.objects.filter(name="Plio 1").first()
```
dalmia, line 148: is it sure that this test case will run before PlioTestCase? why depend on it instead of creating a new plio instance here itself?

dalmia, line 148: we are not using multiple plios in this test case - so, can only keep `plio`

plio/tests.py
```diff
@@ -135,3 +135,106 @@ def setUp(self):
...
+
+        self.item_1_plio_1 = Item.objects.create(
+            plio=self.plio_1, type="question", time=1
+        )
+        self.item_2_plio_1 = Item.objects.create(
+            plio=self.plio_1, type="question", time=2
+        )
+
+        self.question_1_plio_1 = Question.objects.create(item=self.item_1_plio_1)
+        self.question_2_plio_1 = Question.objects.create(item=self.item_2_plio_1)
```
dalmia, line 158: on the basis of the comment above - don't need `plio_1` in the names

plio/tests.py
```diff
@@ -135,3 +135,106 @@ def setUp(self):
...
+        """
+        # upload a test image and retrieve the id
+        with open("plio/static/plio/test_image.jpeg", "rb") as img:
+            response = self.client.post(reverse("images-list"), {"url": img})
+        uploaded_image_id = response.json()["id"]
+
+        # update the question with the newly uploaded image
+        response = self.client.put(
+            reverse("questions-detail", args=[self.question_1_plio_1.id]),
+            {"item": self.item_1_plio_1.id, "image": uploaded_image_id},
```
dalmia, line 172: item does not need to passed in put, no? it only updates whatever is being passed - so, item should remain unchanged if not passed.

plio/tests.py
```diff
@@ -135,3 +135,106 @@ def setUp(self):
...
+        Tests whether an authorized user can create/upload an image
+        """
+        with open("plio/static/plio/test_image.jpeg", "rb") as img:
+            response = self.client.post(reverse("images-list"), {"url": img})
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+    def test_guest_cannot_upload_image(self):
+        """
+        Tests whether a guest(unauthroized user) should not be able to create/upload an image
```
dalmia, line 224:
```suggestion
        Tests whether a guest(unauthorized user) should not be able to create/upload an image
```

plio/tests.py
```diff
@@ -135,3 +135,106 @@ def setUp(self):
...
+        )
+        # the user should be able to update the question
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_user_cannot_attach_image_to_other_user_question(self):
+        """
+        Tests that a user should NOT be able to link images to the questions that were created by some other user
+        """
+        # create a new user
+        new_user = User.objects.create(mobile="+919988776654")
```
dalmia, line 182: Vaibhav used the number `+919876543210` as dummy - best to use the same everywhere?

plio/tests.py
```diff
@@ -135,3 +135,106 @@ def setUp(self):
...
+        new_access_token = AccessToken.objects.create(
+            user=new_user,
+            application=self.application,
+            token=random_token,
+            expires=expires,
+            scope=scopes,
+        )
+        # reset and set the new credentials
+        self.client.credentials()
+        self.client.credentials(HTTP_AUTHORIZATION="Bearer " + new_access_token.token)
```
dalmia, line 198: shouldn't the second line alone do it?

plio/tests.py
deepansh96: Screenshot 2021-06-15 at 8 53 41 PM

plio/models.py
```diff
@@ -35,9 +35,10 @@ def _generate_unique_uuid(self):
 
     def save(self, *args, **kwargs):
         """Image save method. Before uploading, it creates a unique file name for the image."""
-        self.url.name = "".join(
-            [self._generate_unique_uuid(), os.path.splitext(self.url.name)[1]]
-        )
+        if not self.url.name:
```
dalmia, line 38:
```suggestion
        # to avoid setting a new name every time the model is saved
        if not self.url.name:
```

PR #189, opened 2021-06-15
Organization test cases

rathorevaibhav:
## Summary

Some more test cases for organization

## Test Plan

- [x] Wrote tests
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,3 @@ jobs:
 
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
-        with:
-          # The codecov token to upload report. Can be retrieved from your Codecov dashboard.
-          # This should be set from `GitHub Repo > Settings > Secrets`.
-          token: ${{ secrets.CODECOV_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,4 +59,3 @@ jobs:
           # The codecov token to upload report. Can be retrieved from your Codecov dashboard.
           # This should be set from `GitHub Repo > Settings > Secrets`.
           token: ${{ secrets.CODECOV_TOKEN }}
-          fail_ci_if_error: true
--- a/organizations/tests.py
+++ b/organizations/tests.py
@@ -11,8 +11,30 @@ class OrganizationTestCase(BaseTestCase):
         Organization.objects.create(name="Org 1", shortcode="org-1")
         Organization.objects.create(name="Org 2", shortcode="org-2")
 
+    def test_guest_cannot_list_organization(self):
+        # unset the access token so that API requests go as unauthenticated user
+        self.client.credentials()
+        # Make a request without access token
+        response = self.client.get(reverse("organizations-list"))
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_guest_cannot_list_organization_random_token(self):
+        # Random token should give 401_unauthorized status
+        invalid_token = "abcd"
+        self.client.credentials(HTTP_AUTHORIZATION="Bearer " + invalid_token)
+
+        response = self.client.get(reverse("organizations-list"))
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+
     def test_non_superuser_cannot_list_organizations(self):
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -59,4 +59,3 @@ jobs:
           # The codecov token to upload report. Can be retrieved from your Codecov dashboard.
           # This should be set from `GitHub Repo > Settings > Secrets`.
           token: ${{ secrets.CODECOV_TOKEN }}
-          fail_ci_if_error: true
```
dalmia, line 62: any reason for removing this?

organizations/tests.py
```diff
@@ -11,8 +11,30 @@ def setUp(self):
         Organization.objects.create(name="Org 1", shortcode="org-1")
         Organization.objects.create(name="Org 2", shortcode="org-2")
 
+    def test_guest_cannot_list_organization(self):
```
dalmia, line 14: isn't this already tested by `test_non_superuser_cannot_list_organizations`?

organizations/tests.py
```diff
@@ -11,8 +11,30 @@ def setUp(self):
...
     def test_non_superuser_cannot_list_organizations(self):
         """A non-superuser should not be able to list organizations"""
         # get organizations
         response = self.client.get(reverse("organizations-list"))
         self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_superuser_can_list_organizations(self):
+        """A superuser should be able to list organizations"""
+        self.user.is_superuser = True
+        self.user.save()
```
dalmia, line 38: will this save the user as superuser for other tests too or only for this test?

.github/workflows/ci.yml
rathorevaibhav: Yeah, the integration is not 100% perfect so this is giving trouble. I'll look more into this for open source projects. Ideally we don't need any token for open source repos.

Having this in place is kind of false alarm that the test cases didn't succeed.

organizations/tests.py
rathorevaibhav: Non superusers can be valid users (regular users, org-admins etc). Guests are unauthenticated users.

organizations/tests.py
rathorevaibhav: Just this test. The DB resets after every test case and re-run everything that's there in `setUp()`.

.github/workflows/ci.yml
```diff
@@ -55,8 +55,3 @@ jobs:
 
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
-        with:
-          # The codecov token to upload report. Can be retrieved from your Codecov dashboard.
-          # This should be set from `GitHub Repo > Settings > Secrets`.
-          token: ${{ secrets.CODECOV_TOKEN }}
-          fail_ci_if_error: true
```
dalmia, line 62: should we then enable this again?

rathorevaibhav: We don't need token. I think because it was failing before and I wasn't sure of the reason, I thought having token will help. But I see the upload to CodeCov is working fine without token.

On `fail_ci_if_error`, I think we can skip it too. The more important thing is the test cases shouldn't fail. Additionally, now we've removed token and our repo will be treated as open-source, I don't think upload to codecov will fail ever.

plio/tests.py
```diff
@@ -113,10 +113,26 @@ def test_user_cannot_duplicate_other_user_plio(self):
...
+            title="Video 2", url="https://www.youtube.com/watch?v=jWdA2JFCxGw"
+        )
 
-    def test_for_video(self):
-        # write API calls here
-        self.assertTrue(True)
+    def test_guest_cannot_list_videos(self):
+        # unset the credentials
+        self.client.credentials()
+        # get plios
```
dalmia, line 127:
```suggestion
        # get videos
```

plio/tests.py
```diff
@@ -113,10 +113,26 @@ def test_user_cannot_duplicate_other_user_plio(self):
...
+        response = self.client.get(reverse("videos-list"))
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_user_can_list_videos(self):
+        # get plios
```
dalmia, line 132:
```suggestion
        # get videos
```

PR #192, opened 2021-06-17
Download dashboard data from BigQuery

rathorevaibhav:
## Summary

1. Downloads data from BigQuery if BigQuery Enabled env variable is True
2. Docker-compose `platform: linux/amd64` to support python dependency `pyarrow` on Apple Silicon
3. Documentation updates
4. Organization middleware updated to return default `public` schema. Postgres uses public by default but BigQuery doesn't. Hence had to hardcode.
5. Queries updated to support both BigQuery and Postgres

## Test Plan

- [x] Wrote tests
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production


organizations/middleware.py
```diff
@@ -37,7 +37,8 @@ def get_schema(self, request):
         tenant = self.get_tenant(request)
         if tenant:
             return tenant.schema_name
-        return None
+        # as get_schema is being used when querying BigQuery datasets, we explicity need to mention `public`
+        return "public"
```
dalmia, line 41: I think this could be covered by tests easily right?

plio/views.py
```diff
@@ -223,13 +239,35 @@ def download_data(self, request, uuid):
         # schema name to query in
         schema_name = OrganizationTenantMiddleware().get_schema(self.request)
 
+        if BIGQUERY["enabled"]:
+            gcp_service_account_file = "/tmp/gcp-service-account.json"
```
dalmia, line 243: should we keep it as part of the repo as in `plio-analytics` and move it to `.gitignore`? Also, will need to update deployment / installation steps to include info about this file.

organizations/middleware.py
rathorevaibhav: Yeah, this is already being covered as any API calls will be running this middleware.

The only thing that wasn't running was the statements when we have tenant present. They should now also run in one test case.

plio/tests.py
```diff
@@ -268,3 +273,79 @@ def test_each_image_has_unique_name(self):
...
+    def test_non_plio_owner_cannot_download_data(self):
+        # create new user
+        new_user = User.objects.create(mobile="+919988776655")
+        # make a plio with new user
+        new_user_plio = Plio.objects.create(
+            name="Plio 2", video=self.video, created_by=new_user, status="published"
+        )
+        # download new user plio data
+        response = self.client.get(f"/api/v1/plios/{new_user_plio.uuid}/download_data/")
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
```
dalmia, line 316: I think this comes under the other bucket of a non-owner not being able to access the plio itself - which is covered in the plio test case - so, this could be redundant?

rathorevaibhav: Yeah, I had the same thought. But since it's a different view, there may be scenarios in the future where we may have to write separate permissions for plio access and plio download. That's why I thought of adding it here as well.

rathorevaibhav: I think with test cases, the more we cover, it's better. Just so that in future say the plio access functionality changes, we know the areas impacted from failing test cases.

plio/tests.py
```diff
@@ -268,3 +273,79 @@ def test_each_image_has_unique_name(self):
...
+
+    def test_published_plio_data_can_be_downloaded(self):
+        # download plio data
+        response = self.client.get(f"/api/v1/plios/{self.plio.uuid}/download_data/")
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertTrue(isinstance(response, FileResponse))
+
+    def test_non_plio_owner_cannot_download_data(self):
+        # create new user
+        new_user = User.objects.create(mobile="+919988776655")
```
dalmia, line 309: have created a `self.user_2` in my recent change in the `BaseTestCase` merged to PR so that you don't have to create a new user again.

plio/tests.py
```diff
@@ -268,3 +273,79 @@ def test_each_image_has_unique_name(self):
...
+        new_user_plio = Plio.objects.create(
+            name="Plio 2", video=self.video, created_by=new_user, status="published"
+        )
+        # download new user plio data
+        response = self.client.get(f"/api/v1/plios/{new_user_plio.uuid}/download_data/")
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
+
+    def test_user_can_download_data_for_other_user_plio_in_organization(self):
+        # create new user
+        new_user = User.objects.create(mobile="+919988776655")
```
dalmia, line 320: same comment as above.

plio/tests.py
```diff
@@ -268,3 +273,79 @@ def test_each_image_has_unique_name(self):
...
+        # create a new organization
+        organization = Organization.objects.create(name="Org 1", shortcode="org-1")
```
dalmia, line 322: have created `self.organization` in `BaseTestCase` so this can be removed.

plio/tests.py
```diff
@@ -268,3 +273,79 @@ def test_each_image_has_unique_name(self):
...
+
+        # add both users to the organization
+        role = Role.objects.filter(name="org-view").first()
```
dalmia, line 325: have created `self.org_view_role`, so this can be deleted.

plio/tests.py
```diff
@@ -1,18 +1,23 @@
 import datetime
```
dalmia, line 1: do merge in the latest changes from master

PR #197, opened 2021-06-19
Tests for Users App

### LINKED ISSUE #196: More tests for user app

dalmia:
Fixes #196 

## Summary
- tests for `user`, `organization-user` and other functions defined in `users/views.py`
- added analytics env variables to CI for testing the function to fetch analytics app access token
- created a config file for `pep8speaks`

## Test Plan
- [x] Wrote tests
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production
- [x] ~If changes in DB, update [DB schema](https://docs.google.com/presentation/d/15c45HbHphh4w3guaFV3BgEA5nMKd-4J35JdUKdrpPzo/edit#slide=id.gc8c9ec0b5e_0_0)~

### DIFF
--- a/users/tests.py
+++ b/users/tests.py
@@ -45,6 +45,20 @@ class OtpAuthTestCase(BaseTestCase):
         )
         self.assertEqual(response.status_code, status.HTTP_200_OK)
 
+    def test_valid_otp_should_pass_new_user(self):
+        # new user
+        new_user_mobile = "+919988776654"
+
+        # request otp
+        self.client.post(reverse("request-otp"), {"mobile": new_user_mobile})
+
+        # verify valid otp
+        otp = OneTimePassword.objects.filter(mobile=new_user_mobile).first()
+        response = self.client.post(
+            reverse("verify-otp"), {"mobile": new_user_mobile, "otp": otp.otp}
+        )
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
 
 class UserMetaTestCase(BaseTestCase):
     def setUp(self):
--- a/users/tests.py
+++ b/users/tests.py
@@ -1,7 +1,7 @@
 import json
 from rest_framework import status
 from django.urls import reverse
-from users.models import OneTimePassword, User, Role
+from users.models import OneTimePassword, Role, OrganizationUser
 from plio.tests import BaseTestCase
 from organizations.models import Organization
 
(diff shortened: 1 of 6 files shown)


users/tests.py
```diff
@@ -12,39 +13,136 @@ def setUp(self):
         super().setUp()
         # unset client credentials token so that the subsequent API calls goes as guest
         self.client.credentials()
-        self.user_mobile = "+919876543210"
 
     def test_guest_can_request_for_otp(self):
         response = self.client.post(
-            reverse("request-otp"), {"mobile": self.user_mobile}
+            reverse("request-otp"), {"mobile": self.user.mobile}
```
deepansh96, line 19: I can't see `user.mobile` being set in the BaseTestCase
Default is None, is it being set to None here?

users/tests.py
```diff
@@ -12,39 +13,136 @@ def setUp(self):
...
+
+        # get config
+        response = self.client.get(f"/api/v1/users/{self.user.id}/config/")
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.json(), config)
+
+    def test_user_cannot_get_other_user_config(self):
+        # get config
+        response = self.client.get(f"/api/v1/users/{self.user_2.id}/config/")
```
deepansh96, line 78: isnt it required that user 2 should also have a config?
I guess it tries to fetch the default config here right?

users/tests.py
```diff
@@ -12,39 +13,136 @@ def setUp(self):
...
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_update_config_no_config_provided(self):
```
deepansh96, line 81: one liner comment, the function name is not super clear

deepansh96: or can write `test_cannot_update_with_empty_config`

users/tests.py
```diff
@@ -12,39 +13,136 @@ def setUp(self):
...
+        # update config
+        response = self.client.patch(f"/api/v1/users/{self.user.id}/config/", {})
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_update_config_extra_params(self):
```
deepansh96, line 86: one liner comment

users/tests.py
```diff
@@ -67,21 +165,70 @@ def test_for_role(self):
...
+        self.assertEqual(response.json()[0]["user"], self.user.id)
+        self.assertEqual(response.json()[1]["user"], self.user.id)
+        self.assertEqual(response.json()[0]["organization"], self.organization.id)
+        self.assertEqual(response.json()[1]["organization"], self.organization_2.id)
+
+    def test_normal_user_cannot_list_organization_users(self):
+        # get organization users
+        response = self.client.get(reverse("organization-users-list"))
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(response.json()), 0)
```
deepansh96, line 201: This should actually not return 200 :P 
Not allowed or bad request or some error code should come, ideally

users/tests.py
```diff
@@ -67,21 +165,70 @@ def test_for_role(self):
...
+        )
+
+        # get organization users
+        response = self.client.get(reverse("organization-users-list"))
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(response.json()), 2)
+        self.assertEqual(response.json()[0]["user"], self.user.id)
+        self.assertEqual(response.json()[0]["organization"], self.organization.id)
+        self.assertEqual(response.json()[1]["user"], self.user_2.id)
+        self.assertEqual(response.json()[1]["organization"], self.organization.id)
```
deepansh96, line 222: wouldn't it make sense to test if an org admin of org 1 cannot list the org-users of some other org?

deepansh96: This code tests if an org admin can list their org users. This does not test if an org admin can **ONLY** list their org users

deepansh96 (changes requested):

Issue #201, opened 2021-06-21
Monitoring, error logging and reporting


rathorevaibhav:
**Describe the task**
We need to set up monitoring and error logging + reporting for backend setup. This will help with uptime alerts and notifications for errors happening so that we can mimic the user behavior better and optimize our codebase.

**Expected behavior**
The following things need to be worked on:

### Components to consider
1. RDS
2. Fargate instance
3. Elasticache
4. Load balancer
5. CPU utilization + memory consumption threshold hits
6. Auto-scaling triggers

### Metrics to capture for above components
1. Uptime - Production
2. Error reporting - only for production - notification channels to be **Discord** + mail
3. Error logs - for both staging and production
4. Access logs - for both staging and production

### Tools to evaluate
1. Cloudwatch
2. AppSignal
3. Sentry
4. LogRocket
5. Updown.io
6. Datadog
7. Grafana
8. Prometheus

PR #202, opened 2021-06-23
Sentry integration

rathorevaibhav:
Targets application-level monitoring for #201 

## Summary

1. Sentry installation for Django
2. Documentation update

## Test Plan

- [ ] Wrote tests
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production


docs/ERROR-MONITORING.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Pre-requisites
+1. Create an account at Sentry.
+2. Set up your organization.
+3. Create a new Django project. Name it `plio-backend`.
+
+
+### Enable error logging
+1. Set up env variable for `SENTRY_DSN`. You can get the value of DSN for your projects from `Project Settings > SDK Setup > Client Keys (DSN)`
+2. Make sure your `NODE_ENV` is either 'staging' or 'production'.
```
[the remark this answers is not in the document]
rathorevaibhav: Sorry, updated!

docs/ERROR-MONITORING.md
```diff
@@ -0,0 +1,60 @@
...
+    from django.urls import path
+    def trigger_error(request):
+        division_by_zero = 1 / 0
+
+    urlpatterns = [
+        path('trigger-error/', trigger_error),
+        # other paths
+    ]
+    ```
+4. You should now see the error at your Sentry dashboard as well from `Issues` tab.
```
dalmia, line 25: one should only get an error if the trigger-error endpoint is called right?

.env.example
```diff
@@ -71,3 +71,6 @@ EMAIL_HOST_PASSWORD=
 # Use `sns` to have AWS SNS support. The AWS credentials must be present for this.
 # Use an empty string to log SMSs into a file instead. Recommended for development mode.
 SMS_DRIVER='sns'
+
+# Sentry Data Source Name
+SENTRY_DSN=
```
dalmia, line 76: need to update `docs/ENV.md`

docs/ERROR-MONITORING.md
```diff
@@ -0,0 +1,60 @@
...
+### Pre-requisites
+1. Create an account at Sentry.
+2. Set up your organization.
+3. Create a new Django project. Name it `plio-backend`.
+
+
+### Enable error logging
+1. Set up env variable for `SENTRY_DSN`. You can get the value of DSN for your projects from `Project Settings > SDK Setup > Client Keys (DSN)`
+2. Make sure your `APP_ENV` is either 'staging' or 'production'.
+3. Trigger an error by adding the following line in your `urls.py` file:
```
dalmia, line 14:
```suggestion
3. Test your set up by triggering an error - add the following line in your `urls.py` file:
```

docs/ERROR-MONITORING.md
```diff
@@ -0,0 +1,60 @@
...
+        # other paths
+    ]
+    ```
+4. You should now see the error at your Sentry dashboard as well from `Issues` tab.
+
+### Enable reporting
+By default, Sentry sends error reports to your inbox. However, you may prefer other channels to receive error notifications.
+
+At Plio, we use Discord to receive real-time alerts using Sentry Webhooks. If you're also using Discord, or a different tool that supports webhooks, follow the steps below to start receiving alerts:
+1. Create a new channel at Discord where you want to receive the alerts. Then, enable the webhook and copy webhook url. `Channel settings > Integrations > Webhook`
```
dalmia, line 31:
```suggestion
1. Create a new channel at Discord where you want to receive the alerts. Then, enable the webhook and copy the webhook url. `Channel settings > Integrations > Webhook`
```

docs/ERROR-MONITORING.md
```diff
@@ -0,0 +1,60 @@
...
+2. Create a SentryDiscord webhook.
```
dalmia, line 32: this has been done for staging right? I still need to set another one for prod right? And I should be able to test it locally by setting APP_ENV and SENTRY_DSN appropriately?

docs/ERROR-MONITORING.md
dalmia, line 25:
```suggestion
4. You should now see the error at your Sentry dashboard as well from the `Issues` tab.
```

rathorevaibhav: Yes. But you may receive errors on existing endpoints also if there are any. The code covers all endpoints.

.env.example
rathorevaibhav: Man! Every time I think I didn't miss anything! Updating...

plio/settings.py
```diff
@@ -322,3 +322,22 @@
...
+    # Refer Sentry documentation for more configs: https://docs.sentry.io/platforms/python/configuration/
+    sentry_sdk.init(
+        dsn=SENTRY_DSN,
+        integrations=[DjangoIntegration()],
+        # Set traces_sample_rate to 1.0 to capture 100%
+        # of transactions for performance monitoring.
+        traces_sample_rate=1.0,
+        # If you wish to associate users to errors (assuming you are using
+        # django.contrib.auth) you may enable sending PII data.
+        send_default_pii=True,
```
dalmia, line 341: exactly which params are passed?

docs/ERROR-MONITORING.md
rathorevaibhav: This hasn't been done for staging or production as the code deployment was pending.

Yes, you can test it by setting APP_ENV and SENTRY_DSN.

plio/settings.py
rathorevaibhav: @dalmia it seems like we may want to remove this option. While it gives more info on the error, we may be violating user privacy.

This link in the documentation gives more idea about what PII includes:
https://docs.sentry.io/platforms/python/guides/django/data-management/sensitive-data/

What do you think?

rathorevaibhav: Initially, as it was in the example snippet I was using, it felt like a good thing. But upon reading more, it seems either we need to find a middle ground or not use it.

Maybe we can go without this option and see if we're having difficulty replicating issues. If not, then we're good without it. If yes, we can check up to what level of PII we need and how to configure that.

deepansh96: Maybe let's just send the user's DB id and scrub everything else?

rathorevaibhav: By default it sends user id, email and username. Refer function `_set_user_info` in this file below:
https://github.com/getsentry/sentry-python/blob/b0658904925ec2b625b367ae86f9762b5a382d5f/sentry_sdk/integrations/django/__init__.py#L451


I've added a `strip_sensitive_data` function that removes email and username.

plio/settings.py
```diff
@@ -329,15 +329,27 @@
     import sentry_sdk
     from sentry_sdk.integrations.django import DjangoIntegration
 
+    def strip_sensitive_data(event, hint):
+        """Strips user email and username from the event/error details. Only data remains is event['user']['id']."""
+        try:
+            event["user"].pop("email")
+            event["user"].pop("username")
```
dalmia, line 336: there is no username field for us right? 
We should also add `mobile` here right?

rathorevaibhav: Yes. it's not there but it's set by default with `django.contrib.auth`. Just as a precaution, I've deliberately popped that element. If it's not present, the try/catch block will take care.

docs/ENV.md
```diff
@@ -146,10 +146,16 @@ The BigQuery project id that contains the datasets.
...
 Once you have downloaded the JSON file, run the following commands and use the output for this environment variable:
 ```sh
 cat /path/to/gcp-service-account-filename.json | base64
 ```
+
+### Error monitoring
+Plio supports error monitoring on your app with [Sentry](https://sentry.io/). Visit our guide on [Error Monitoring](./docs/../ERROR-MONITORING.md) to enable it for your Plio setup.
+
+#### `SENTRY_DSN`
+The Sentry Data Source Name. This is a unique link that identifies your Sentry project and sends error reports to Sentry dashboard. You can get the value of DSN for your projects from `Sentry dashboard > Project Settings > SDK Setup > Client Keys (DSN)`.
```
deepansh96, line 161:
```suggestion
The Sentry Data Source Name. This is a unique link that identifies your Sentry project and sends error reports to Sentry dashboard. You can get the value of DSN for your projects from `Sentry dashboard > Settings > Project > {PROJECT_NAME} > SDK Setup > Client Keys (DSN)`.
```

deepansh96: It took me minute to find this on Sentry :p

deepansh96 (changes requested):

PR #207, opened 2021-06-23
Avanti auth override support

### LINKED ISSUE #206: Support authentication for plio using third party auth tokens

deepansh96:
Fixes #206
Depends on [this PR](https://github.com/avantifellows/plio-frontend/pull/365)

## Test Plan

- [ ] Wrote tests
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production


users/config.py
```diff
@@ -1 +1,5 @@
 user_status_choices = [("waitlist", "Added to Waitlist"), ("approved", "Approved")]
+auth_type_choices = [
+    ("default", "Default Authentication"),
```
dalmia, line 3: I think we can keep default as None right? - we don't need this choice? - and allow it to be nullable?

users/models.py
```diff
@@ -71,6 +75,10 @@ class User(SafeDeleteModel, AbstractUser):
     status = models.CharField(
         max_length=255, choices=user_status_choices, default="approved"
     )
+    auth_type = models.CharField(
+        max_length=255, choices=auth_type_choices, default="default"
```
dalmia, line 79: same comment as above

users/models.py
```diff
@@ -15,6 +15,8 @@ def create_user(
         is_admin=False,
         is_staff=False,
         is_active=True,
+        auth_type="default",
```
dalmia, line 18: same comment as above

users/views.py
```diff
@@ -211,6 +217,55 @@ def get_by_access_token(request):
...
+            {"detail": "auth_type not provided"}, status=status.HTTP_400_BAD_REQUEST
+        )
+    if "unique_id" not in request.data:
+        return Response(
+            {"detail": "unique_id not provided"}, status=status.HTTP_400_BAD_REQUEST
+        )
+    if "access_token" not in request.data:
+        return Response(
+            {"detail": "access_token not provided"}, status=status.HTTP_400_BAD_REQUEST
+        )
```
dalmia, line 237: this can be done via a for loop over the keys?

users/views.py
```diff
@@ -211,6 +217,55 @@ def get_by_access_token(request):
...
+        return Response(
+            {"detail": "access_token invalid"}, status=status.HTTP_401_UNAUTHORIZED
+        )
+
+    user = User.objects.filter(
+        auth_type=request.data["auth_type"],
+        unique_id=request.data["unique_id"],
+    ).first()
+
+    if not user:
```
dalmia, line 252: comment

users/views.py
```diff
@@ -211,6 +217,55 @@ def get_by_access_token(request):
...
+        )
+
+    # define the backend authenticator
+    user.backend = "oauth2_provider.contrib.rest_framework.OAuth2Authentication"
+    login(request, user)
+
+    # define an application
+    application = Application.objects.get(name=API_APPLICATION_NAME)
+    # get the newly generated access/refresh tokens for the user and application
+    token = get_new_access_token(user, application)
```
dalmia, line 264: this part also seems same as verify_otp - can be made a function?

users/tests.py
```diff
@@ -1,9 +1,10 @@
 import json
 from rest_framework import status
 from django.urls import reverse
-from users.models import OneTimePassword, Role, OrganizationUser
+from users.models import OneTimePassword, Role, OrganizationUser, User
 from plio.tests import BaseTestCase
 from organizations.models import Organization
+from oauth2_provider.models import AccessToken, RefreshToken
```
dalmia, line 7: installed modules should be imported before app's modules.

users/tests.py
```diff
@@ -60,6 +61,51 @@ def test_valid_otp_should_pass_new_user(self):
...
+    def test_cannot_authenticate_without_all_required_keys(self):
+        # keeping only 2 out of the 3 required auth keys
+        payload = {"auth_type": "abcd", "unique_id": "1234"}
+        response = self.client.post(
+            reverse("convert_third_party_access_token"), payload
+        )
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.data["detail"], "access_token not provided")
+
+    def test_guest_can_authenticate_using_third_party(self):
```
dalmia, line 81: should then be one more test to ensure that if someone was already logged in, and then this third party auth is sent, then the authenticated user is changed.

dalmia, line 81: one more test for user authenticating through 3rd party when that user already exists.

users/tests.py
```diff
@@ -60,6 +61,51 @@ def test_valid_otp_should_pass_new_user(self):
...
+        refresh_token_exists = RefreshToken.objects.filter(
+            token=response.data["refresh_token"]
+        ).exists()
+        user_created = User.objects.filter(
+            auth_type=third_party_auth_details["auth_type"],
+            unique_id=third_party_auth_details["unique_id"],
+        ).exists()
+        self.assertTrue(access_token_exists)
+        self.assertTrue(refresh_token_exists)
+        self.assertTrue(user_created)
```
dalmia, line 106: one more check that current user has been set to this new user

users/tests.py
deepansh96: Hard to test here. Initially we can check ki `self.client` mein jo HTTP_AUTH header set hai, that is of some user_1 (which was set up in the `setUp` method of the `BaseTestClass`)

But the problem is, when we authenticate with a third party auth, it just gives us the access token. "Currently logged in user" jaisa kuch hai nhi isme. That part, frontend takes care of. Frontend mein we could've tested that the currently logged in user has changed. How will we test for the "currently logged in user" here?

users/tests.py
deepansh96: How to check for the current user?

users/tests.py
```diff
@@ -105,6 +107,60 @@ def test_guest_can_authenticate_using_third_party(self):
...
+        # retrieve the currently logged in user from client's credentials
+        current_access_token = self.client._credentials["HTTP_AUTHORIZATION"].split(
+            " "
+        )[1]
+        current_logged_in_user = (
+            AccessToken.objects.filter(token=current_access_token).first().user_id
+        )
+
+        # login a new user via third party authentication
+        self.client.credentials()
```
dalmia, line 120: it should work without unsetting this as well right?

users/tests.py
```diff
@@ -60,6 +62,106 @@ def test_valid_otp_should_pass_new_user(self):
...
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.data["detail"], "access_token not provided")
+
+    def test_guest_can_authenticate_using_third_party(self):
+        # unset client credentials token so that the subsequent API calls goes as guest
+        self.client.credentials()
+
+        # some dummy third party auth details
+        third_party_auth_details = {
+            "auth_type": "abcd",
```
dalmia, line 86: auth_type = `abcd` should be invalid no?

users/tests.py
```diff
@@ -60,6 +62,106 @@ def test_valid_otp_should_pass_new_user(self):
...
+        self.assertNotEqual(current_logged_in_user, new_logged_in_user)
+
+    def test_existing_third_party_user_can_authenticate_again(self):
+        # create a third party user
+        user = User.objects.create(auth_type="test_auth", unique_id="test_id")
+        total_users = User.objects.count()
+
+        # authenticate the same user again via third party auth
+        third_party_auth_details = {
+            "auth_type": "test_auth",
```
dalmia, line 145: let's use same value for testing

users/tests.py
```diff
@@ -60,6 +62,106 @@ def test_valid_otp_should_pass_new_user(self):
...
+            " "
+        )[1]
+        current_logged_in_user = (
+            AccessToken.objects.filter(token=current_access_token).first().user_id
+        )
+
+        # login a new user via third party authentication
+        self.client.credentials()
+        third_party_auth_details = {
+            "auth_type": "abcd",
```
dalmia, line 122: same as above

users/views.py
```diff
@@ -228,18 +228,32 @@ def convert_third_party_access_token(request):
...
+    # `auth_type` should be one of the values from `auth_type_choices`
+    if (
+        len(
+            [
+                choice
+                for choice in auth_type_choices
+                if choice[0] == request.data["auth_type"]
+            ]
+        )
+        == 0
```
dalmia, line 247: `not len`

PR #210, opened 2021-06-29
FIX: refactored third party auth according to new logic

### LINKED ISSUE #209: Refactor the code according to the new third party auth logic (using API_KEYS)
https://whimsical.com/plio-docs-structure-SVJKqZkq2r1bV6Sz4kk8iZ@3CRerdhrAwxbw9WtHJH555kP

deepansh96:
Fixes #209 

## Test Plan

- [x] Wrote tests
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production


users/views.py
```diff
@@ -224,9 +225,9 @@ def get_by_access_token(request):
 
 @api_view(["POST"])
 @permission_classes([AllowAny])
-def convert_third_party_access_token(request):
+def convert_api_key_to_token(request):
     """
-    Convert any third party auth token into Plio's internal access token
+    Convert an api_key that Plio created for a third party org, into Plio's internal access token
```
dalmia, line 230:
```suggestion
    Convert an API Key that Plio created for a third party org into Plio's internal access token
```

users/views.py
```diff
@@ -235,36 +236,28 @@ def convert_third_party_access_token(request):
...
 
-    # TODO - verification step - allowed for now
-    token_verified = False
-
-    if not token_verified:
-        return Response(
-            {"detail": "access_token invalid."}, status=status.HTTP_400_BAD_REQUEST
-        )
+    # the org to which this api_key belongs to
+    requested_org = api_key.first()
```
dalmia, line 248:
```suggestion
    requesting_org = api_key.first()
```

users/serializers.py
```diff
@@ -24,8 +24,8 @@ class Meta:
             "updated_at",
             "organizations",
             "status",
-            "auth_type",
             "unique_id",
+            "org",
```
dalmia, line 28: I think clearer would be `auth_org`.

users/models.py
```diff
@@ -75,8 +75,13 @@ class User(SafeDeleteModel, AbstractUser):
     status = models.CharField(
         max_length=255, choices=user_status_choices, default="approved"
     )
-    auth_type = models.CharField(max_length=255, choices=auth_type_choices, null=True)
     unique_id = models.CharField(max_length=255, null=True)
+    org = models.ForeignKey(
+        Organization, on_delete=models.DO_NOTHING, 
+        related_name="org",
```
dalmia, line 81: why use this field?

plio/urls.py
```diff
@@ -27,7 +27,7 @@
     verify_otp,
     get_by_access_token,
     retrieve_analytics_app_access_token,
-    convert_third_party_access_token,
+    convert_api_key_to_token,
```
dalmia, line 30: should probably change this endpoint name - something like `get_external_auth_access_token`

organizations/tests.py
```diff
@@ -5,12 +5,16 @@
...
-        super().setUp()
+    @classmethod
+    def setUpTestData(self):
+        super().setUpTestData()
         # seed some organizations
         Organization.objects.create(name="Org 1", shortcode="org-1")
         Organization.objects.create(name="Org 2", shortcode="org-2")
 
+    def setUp(self):
+        super().setUp()
```
dalmia, line 16: this is not needed. If you don't define a function in an inherited class, it would do this by default.

organizations/models.py
```diff
@@ -18,6 +20,15 @@ class Organization(TenantMixin, SafeDeleteModel):
     class Meta:
         db_table = "organization"
 
+    def _generate_random_cryptographically_secure_string(self, length=20):
```
dalmia, line 23: crypto word can come in the description.
```suggestion
    def _generate_random_secure_string(self, length=20):
```

organizations/models.py
```diff
@@ -29,10 +40,18 @@ def _generate_unique_schema_name(self):
...
+        api_key = self._generate_random_cryptographically_secure_string()
+        while Organization.objects.filter(api_key=api_key).exists():
+            api_key = self._generate_random_cryptographically_secure_string()
+        return api_key
+
     def save(self, *args, **kwargs):
         """Organization save method. Before checking it creates a unique schema name if does not exist already."""
         if not self.schema_name:
             self.schema_name = self._generate_unique_schema_name()
+            self.api_key = self._generate_unique_api_key()
```
dalmia, line 54: this should be preceeded by `if not self.api_key` - we should not mix it with schema name, no?

organizations/models.py
```diff
@@ -29,10 +40,18 @@ def _generate_unique_schema_name(self):
             schema_name = self._generate_random_string()
         return schema_name
 
+    def _generate_unique_api_key(self):
+        """Generates a unique api_key."""
+        api_key = self._generate_random_cryptographically_secure_string()
+        while Organization.objects.filter(api_key=api_key).exists():
+            api_key = self._generate_random_cryptographically_secure_string()
```
dalmia, line 47: this line is not covered in tests.

users/views.py
dalmia, line 230: thoda aur clear sochna hoga ismein kuch - API key to access token actually nahi kar rahe na? unique ID + API Key ko kar rahe - and API key can also be used for other things beyond auth - shayad end points ko bhi rename karna chaiye - something like: `get_external_auth_access_token` type.

users/tests.py
```diff
@@ -72,120 +70,97 @@ def setUp(self):
         super().setUp()
 
     def test_cannot_authenticate_without_all_required_keys(self):
-        # keeping only 2 out of the 3 required auth keys
-        payload = {"auth_type": "avanti", "unique_id": "test_id"}
-        response = self.client.post(
-            reverse("convert_third_party_access_token"), payload
-        )
+        # not sending all the required auth keys
```
dalmia, line 73: should check both the versions - any of the keys missed.

users/models.py
deepansh96: This is the error I get if I don't use the `related_name` field
```bash
ERRORS:
users.User.org_id: (fields.E304) Reverse accessor for 'User.org_id' clashes with reverse accessor for 'User.organizations'.
	HINT: Add or change a related_name argument to the definition for 'User.org_id' or 'User.organizations'.
users.User.organizations: (fields.E304) Reverse accessor for 'User.organizations' clashes with reverse accessor for 'User.org_id'.
	HINT: Add or change a related_name argument to the definition for 'User.organizations' or 'User.org_id'.
```
I'm guessing this is because we already have a ManyToMany relationship defined with `Organization`. So if we don't give a `related_name` it gets confused between the two.

organizations/models.py
deepansh96: There is no way to test this
Screenshot 2021-06-30 at 1 01 34 PM

PR #215, opened 2021-07-02
More tests

### LINKED ISSUE #214: Test cases for Sessions and Plio Ordering

deepansh96:
Fixes #214 

## Test Plan

- [x] Wrote tests
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production
- [ ] ~If changes in DB, update [DB schema](https://docs.google.com/presentation/d/15c45HbHphh4w3guaFV3BgEA5nMKd-4J35JdUKdrpPzo/edit#slide=id.gc8c9ec0b5e_0_0)~

### DIFF
--- a/entries/serializers.py
+++ b/entries/serializers.py
@@ -27,13 +27,41 @@ class SessionSerializer(serializers.ModelSerializer):
             "updated_at",
         ]
 
-    def validate_plio(self, plio):
-        """Ensure that a session is created only for a published plio"""
-        if plio.status == "draft":
+    def validate(self, data):
+        """
+        All validation checks happen in this method. Currently these checks are defined
+        - A session can only be created for a published plio
+        - `is_first` is set properly depending on if a session already exists or not
+        """
+        # import ipdb; ipdb.set_trace()
+        # validate plio's status
+        if data["plio"].status == "draft":
             raise serializers.ValidationError(
                 "A session can only be created for a published plio"
             )
-        return plio
+
+        # set `is_first` properly
+        last_session = (
+            Session.objects.filter(plio_id=data["plio"].id)
+            .filter(user_id=data["user"].id)
+            .first()
+        )
+
+        if self.context["view"].action == "create":
+            # while creating a session, check if any session already exists
+            if last_session:
+                data["is_first"] = False
+            else:
+                data["is_first"] = True
+        elif self.context["view"].action == "update":
+            # while updating a session, check if the session being updated
+            # is the very first session
+            if f"{last_session.id}" == self.context["view"].kwargs["pk"]:
+                data["is_first"] = True
+            else:
+                data["is_first"] = False
+
+        return data
 
     def create(self, validated_data):
         """
@@ -58,9 +86,6 @@ class SessionSerializer(serializers.ModelSerializer):
             for key in keys_to_check:
                 if key not in validated_data:
                     validated_data[key] = last_session_data[key]
-
-        else:
-            validated_data["is_first"] = True
 
         # get the newly created session object
         session = Session.objects.create(**validated_data)
--- a/entries/serializers.py
+++ b/entries/serializers.py
@@ -33,7 +33,6 @@ class SessionSerializer(serializers.ModelSerializer):
         - A session can only be created for a published plio
         - `is_first` is set properly depending on if a session already exists or not
         """
-        # import ipdb; ipdb.set_trace()
         # validate plio's status
         if data["plio"].status == "draft":
             raise serializers.ValidationError(
--- a/entries/tests.py
+++ b/entries/tests.py
@@ -1,13 +1,141 @@
+import json
+from django.urls import reverse
+from rest_framework import status
+
 from plio.tests import BaseTestCase
+from plio.models import Plio, Video, Item
 
 
 class SessionTestCase(BaseTestCase):
     def setUp(self):
         super().setUp()
+        # seed a video
+        self.video = Video.objects.create(
+            title="Video 1",
+            url="https://www.youtube.com/watch?v=vnISjBbrMUM",
+            duration=10,
+        )
+        # seed some plios
+        self.plio_1 = Plio.objects.create(
+            name="Plio 1", video=self.video, created_by=self.user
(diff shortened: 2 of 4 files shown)


plio/tests.py
```diff
@@ -288,6 +288,127 @@ def test_duplicate(self):
...
+    def test_default_ordering_when_no_ordering_specified(self):
+        # create a third plio
+        Plio.objects.create(name="Plio 3", video=self.video, created_by=self.user)
+
+        # make a request to list the plio uuids without specifying any order
+        # NOTE: default ordering should come out to be '-updated_at'
+        # order of plios should be [plio_3, plio_2, plio_1]
+        response = self.client.get("/api/v1/plios/list_uuid/")
+        self.assertListEqual(
+            [plio.uuid for plio in Plio.objects.order_by("-updated_at")],
```
dalmia, line 300: I think we should specifically check 3,2,1 here and 1,2,3 there - or else, there could be an error in both Plio.objects.order_by and response.data - and so, both of them are equal but not equal to the value that should be.

plio/tests.py
```diff
@@ -288,6 +288,127 @@ def test_duplicate(self):
...
+        self.plio_2.name = "B_plio"
+        self.plio_2.save()
+        plio_3.name = "C_plio"
+        plio_3.save()
+
+        # `list_uuid` should give the result ordered as [plio_1, plio_2, plio_3]
+        # when "name" ordering is specified
+        response = self.client.get("/api/v1/plios/list_uuid/", {"ordering": "name"})
+        self.assertListEqual(
+            [plio.uuid for plio in Plio.objects.order_by("name")],
```
dalmia, line 336: same comment as above.

entries/tests.py
```diff
@@ -1,13 +1,141 @@
...
+        self.video = Video.objects.create(
+            title="Video 1",
+            url="https://www.youtube.com/watch?v=vnISjBbrMUM",
+            duration=10,
+        )
+        # seed some plios
+        self.plio_1 = Plio.objects.create(
+            name="Plio 1", video=self.video, created_by=self.user
+        )
+        self.plio_2 = Plio.objects.create(
```
dalmia, line 22: not used.

entries/tests.py
```diff
@@ -1,13 +1,141 @@
...
+            name="Plio 2", video=self.video, created_by=self.user
+        )
 
-    def test_for_session(self):
-        # write API calls here
-        self.assertTrue(True)
+        # seed two items linked to plio_1
+        self.item_1 = Item.objects.create(type="question", plio=self.plio_1, time=1)
+        self.item_2 = Item.objects.create(type="question", plio=self.plio_1, time=2)
```
dalmia, line 28: only used once - better to define in the function where they were used.

entries/tests.py
```diff
@@ -1,13 +1,141 @@
...
+            reverse("sessions-detail", args=[session_1["id"]]),
+            {
+                "plio": self.plio_1.id,
+                "watch_time": 5.00,
+                "retention": "1,1,1,1,1,0,0,0,0,0",
+            },
+        )
+        session_1_updated = response.data
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertFalse(session_1_updated["is_first"])
```
dalmia, line 89: why should this change - this should still be true for the current session even after it's updated - it's still the first session.

entries/tests.py
```diff
@@ -1,13 +1,141 @@
...
+        self.assertEqual(session_1_updated["id"], session_1["id"])
```
dalmia, line 90: this seems like a separate test in itself - whether session can be updated.

entries/tests.py
```diff
@@ -1,13 +1,141 @@
...
+
+        # mocking reloading of the page, hence creating a new session
```
dalmia, line 92: should not be mentioning reload etc - that is how it is happening in the frontend - no need of mentioning it in the backend API.

entries/tests.py
```diff
@@ -1,13 +1,141 @@
...
+        # details of the older sessions should be passed along to this session
```
dalmia, line 93:
```suggestion
        # details of the last session should be passed along to this session
```

entries/tests.py
```diff
@@ -1,13 +1,141 @@
...
+        )
+        # after answering a question, update the session
+        response = self.client.put(
+            reverse("sessions-detail", args=[session_2["id"]]),
+            {
+                "plio": self.plio_1.id,
+                "watch_time": 5.00,
+                "retention": "1,1,1,1,1,0,0,0,0,0",
+            },
+        )
```
dalmia, line 123: why is this needed?

entries/tests.py
```diff
@@ -1,13 +1,141 @@
...
+        session_2_updated = response.data
+
+        # mocking reloading again, a new session is created,
```
dalmia, line 126: same comment.

entries/serializers.py
```diff
@@ -27,13 +27,41 @@ class Meta:
...
-    def validate_plio(self, plio):
-        """Ensure that a session is created only for a published plio"""
-        if plio.status == "draft":
+    def validate(self, data):
+        """
+        All validation checks happen in this method. Currently these checks are defined
+        - A session can only be created for a published plio
+        - `is_first` is set properly depending on if a session already exists or not
+        """
+        # import ipdb; ipdb.set_trace()
```
dalmia, line 36: debug line

entries/serializers.py
```diff
@@ -27,13 +27,41 @@ class Meta:
...
+        if self.context["view"].action == "create":
+            # while creating a session, check if any session already exists
+            if last_session:
+                data["is_first"] = False
+            else:
+                data["is_first"] = True
+        elif self.context["view"].action == "update":
+            # while updating a session, check if the session being updated
+            # is the very first session
+            if f"{last_session.id}" == self.context["view"].kwargs["pk"]:
```
dalmia, line 59: can we use `id` instead of `pk` - just like we changed `pk` with `plio` to `uuid`- better to just check `id` if it is available?

PR #231, opened 2021-08-27
Query profiling using silk

rathorevaibhav:
## Summary

1. Added [django-silk](https://github.com/jazzband/django-silk) to the project apps. This will help determine query load times and optimize the system further, especially for load testing.
2. Added some configurations in settings.py to not enable it for production as it has some impact on every request time.

## Test Plan

- [ ] Wrote tests
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production

### DIFF
--- a/plio/settings.py
+++ b/plio/settings.py
@@ -357,9 +357,7 @@ if APP_ENV in ["staging", "production"] and SENTRY_DSN is not None:
         environment=APP_ENV,
     )
 
-# silk query profiling settings
+# settings for django-silk query profiling
 SILKY_AUTHENTICATION = True  # User must login
 SILKY_AUTHORISATION = True  # User must have permissions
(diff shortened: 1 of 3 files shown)


plio/settings.py
```diff
@@ -353,3 +356,8 @@ def strip_sensitive_data(event, hint):
         send_default_pii=True,
         environment=APP_ENV,
     )
+
+# settings for django-silk query profiling
+SILKY_AUTHENTICATION = True  # User must login
+SILKY_AUTHORISATION = True  # User must have permissions
+SILKY_INTERCEPT_PERCENT = 100 if APP_ENV in ["local", "staging"] else 0
```
dalmia, line 363: this would mean that we are intercepting 100% of queries in staging and local but not on production right?

dalmia (approved):

Issue #224, opened 2021-08-24
Use redis cache (django-redis) to cache GET requests


deepansh96:
Instead of querying and pulling from the DB everytime a plio's data is requested, we should use a cache to store all those details in the memory for faster accessing of the data.

This can include plio GET, items GET, questions GET etc

If some information is not cached, we can retrieve it from the DB and store it in the cache for future use

PR #246, opened 2021-09-16
Caching

### LINKED ISSUE #235: Enable Django cache
**Describe the task**
For several DB queries and API responses, we can leverage cache to make the API work faster.

**Expected behavior**
- [ ] Utilize Django's cache framework (using Memcache would be best option) https://docs.djangoproject.com/en/3.2/topics/cache/

rathorevaibhav:
Fixes #235, #224

## Summary

1. Using Django cache with Django-Redis
2. A central cache file that helps with calculating cache keys and invalidating cache
2. Documentation update for caching

## Test Plan

- [ ] Wrote tests
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production


plio/cache.py
```diff
@@ -0,0 +1,34 @@
+from django.core.cache import cache
+
+
+def get_cache_key(instance):
+    """Calculates cache key for an instance"""
+    instance_class = instance.__class__.__name__
+    return {
+        "Video": f"video_{instance.pk}",
```
pritamps, line 8: Upper case is the format you guys have decided?

plio/cache.py
```diff
@@ -0,0 +1,34 @@
+from django.core.cache import cache
+
+
+def get_cache_key(instance):
+    """Calculates cache key for an instance"""
+    instance_class = instance.__class__.__name__
+    return {
```
pritamps, line 7: First time I'm seeing `django-redis`, so maybe this is a dumb question. But is this same method used for all cache hits? So if it's a plio, then other keys will be empty and plio will be the one with a value?

requirements.txt
```diff
@@ -1,42 +1,121 @@
...
+django-request-logging==0.7.2
+django-rest-framework-social-oauth2==1.1.0
+django-s3-storage==0.13.4
+django-safedelete==1.0.0
+django-silk==4.1.0
+django-storages==1.11.1
+django-tenants==3.2.1
+django-utils-six==2.0
+djangorestframework==3.12.2
+drf-cached-instances==0.3.4
```
pritamps, line 43: Lots of new dependencies? :O

pritamps (commented):
Just some minor comments.

Will the cache automatically be updated when a plio is updated? because if not, a GET after an update will return the old value, no? 

Same for deletion. When a plio is deleted, will cache automatically delete it as well? Maybe Django handles all this automatically, just curious :)

plio/cache.py
rathorevaibhav: this is like a switch-case. I'm matching the class name for the instance and using it to determine the key. For a video instance, the class name is `Video` defined inside `plios/models.py`, hence it needs to match.

Do you think it'll be better to transform instance_class to lowercase and then match?

plio/cache.py
rathorevaibhav: Yes, it's switch/case functionality. So it matches the plio class with the keys and returns the corresponding value (which is the cache key name).

For plio instance ID:1, it will return plio_1
For video instance ID:99, it will return video_99

requirements.txt
rathorevaibhav: Not really. I ran `pip freeze > requirements.txt` and it loaded the sub-dependencies also to this file and then sorted them.

It's more or less same. Let me revert. I think it'll be difficult if we need to clear sub-dependencies when we're removing any dependency.

dalmia: @rathorevaibhav use this: https://github.com/bndr/pipreqs

this will not list the sub-dependencies.

docs/CACHING.md
```diff
@@ -0,0 +1,23 @@
...
+
+This guide aims to provide details on how Plio is using caching and pre-requisites for someone contributing to the code.
+
+### Caching workflow
+Plio's caching mechanism can be explained in three simple steps:
+1. Receives an API request
+2. Searches for the requested data in the cache
+3. If not present, query the database and save the response in cache for future requests
+
+![Overview of caching](images/caching-workflow.png)
```
dalmia, line 12: One thing is not clear:
- what happens if the value is present in the cache but has been updated since it was put in the cache?

docs/CACHING.md
```diff
@@ -0,0 +1,23 @@
...
+
+
+### Cache keys
+The calculation of the cache keys are based on the model instances. For example, an instance for plio ID: 1 will have `plio_1` as the cache key.
+For more details, check out the `get_cache_key` function in `plio/cache.py`.
+
+
+### Cache refresh or invalidation
+When a particular instance is updated, it's cached value gets deleted along with any other related instances cache that depends on this cache. For example, if a session instance cache uses plio instance cache in it's response, if the plio is modified, both plio instance and session instance cache will be deleted.
```
dalmia, line 21: Just wording it a bit more clearly.

```suggestion
When a particular instance is updated, its cached value gets deleted along with the cache of any other related instances that depends on this cache. For example, consider a session instance cache that uses a plio instance cache. Now, if the plio is modified, the caches for both the plio instance and the session instance will be deleted.
```

docs/CACHING.md
```diff
@@ -0,0 +1,23 @@
...
+
+The new cache will be set during the first fresh response calculated from the database and will be used for subsequent requests.
```
dalmia, line 23:
```suggestion
The new cache will be set when the first fresh response is calculated from the database and will be used for subsequent requests.
```

users/serializers.py
```diff
@@ -31,10 +33,18 @@ class Meta:
         read_only_fields = ["is_superuser", "is_staff"]
 
     def to_representation(self, instance):
+        # check if a cached version exists and if it does, return it as the response
+        cache_key = get_cache_key(instance)
+        cachedResponse = cache.get(cache_key)
```
dalmia, line 38:
```suggestion
        cached_response = cache.get(cache_key)
```

users/serializers.py
```diff
@@ -31,10 +33,18 @@ class Meta:
...
+        if cachedResponse:
+            return cachedResponse
```
dalmia, line 40:
```suggestion
        cached_response = cache.get(cache_key)
        if cached_response:
            return cached_response
```

users/models.py
```diff
@@ -161,3 +164,9 @@ class OneTimePassword(models.Model):
 
     class Meta:
         db_table = "one_time_password"
+
+
+@receiver(post_save, sender=User)
+@receiver(post_delete, sender=User)
+def user_update_cache(sender, instance, created, raw, **kwargs):
```
dalmia, line 171:
```suggestion
def user_update_cache(_, instance, _, _, **kwargs):
```

plio/models.py
```diff
@@ -140,3 +143,18 @@ class Question(SafeDeleteModel):
...
+    invalidate_cache_for_instance(instance)
+
+
+@receiver(post_save, sender=Video)
+@receiver(post_delete, sender=Video)
+def video_update_cache(sender, instance, **kwargs):
+    # fetch all plio with video id
+    plios = Plio.objects.filter(video_id=instance.id)
+    # invalidate saved cache for the plios
+    invalidate_cache_for_instances(plios)
```
dalmia, line 160: need to do this for question, item, imagw as well no?

plio/cache.py
```diff
@@ -0,0 +1,34 @@
...
+        "Video": f"video_{instance.pk}",
+        "Plio": f"plio_{instance.pk}",
+        "Session": f"session_{instance.pk}",
+        "User": f"user_{instance.pk}",
+        "UserMeta": f"user_meta_{instance.pk}",
+    }.get(instance_class, None)
```
dalmia, line 13: this needs to include question, image and item too as well right?

Why are we including session here? as in, unlike a plio, once a session has been created, changing anything in the plio won't change anything about the session right?

plio/cache.py
```diff
@@ -0,0 +1,34 @@
+from django.core.cache import cache
+
+
+def get_cache_key(instance):
+    """Calculates cache key for an instance"""
```
dalmia, line 5:
```suggestion
    """Calculates the cache key for an instance"""
```

plio/cache.py
```diff
@@ -0,0 +1,34 @@
...
+        "User": f"user_{instance.pk}",
+        "UserMeta": f"user_meta_{instance.pk}",
+    }.get(instance_class, None)
+
+
+def invalidate_cache_for_instance(instance):
+    """Deletes cache for a particular instance"""
+    cache_key = get_cache_key(instance)
+    if cache_key:
+        cache.delete(cache_key)
```
dalmia, line 20: won't this throw an error if the key is not in the cache?

plio/cache.py
```diff
@@ -0,0 +1,34 @@
...
+    keys = []
+    for instance in instances:
+        keys.append(get_cache_key(instance))
+    return keys
+
+
+def invalidate_cache_for_instances(instances):
+    """Deletes cache for a list of instances"""
+    cache_keys = get_cache_keys(instances)
+    cache.delete_many(cache_keys)
```
dalmia, line 34: won't this throw an error if the key is not in the cache?

plio/serializers.py
```diff
@@ -52,12 +54,19 @@ class Meta:
         read_only_fields = ["uuid"]
 
     def to_representation(self, instance):
+        # check if a cached version exists and if it does, return it as the response
+        cache_key = get_cache_key(instance)
+        cachedResponse = cache.get(cache_key)
+        if cachedResponse:
+            return cachedResponse
+
```
dalmia, line 62:
```suggestion
        cached_response = cache.get(cache_key)
        if cached_response:
            return cached_response

```

docs/CACHING.md
```diff
@@ -0,0 +1,23 @@
...
+This guide aims to provide details on how Plio is using caching and pre-requisites for someone contributing to the code.
+
+### Caching workflow
+Plio's caching mechanism can be explained in three simple steps:
+1. Receives an API request
+2. Searches for the requested data in the cache
+3. If not present, query the database and save the response in cache for future requests
+
+![Overview of caching](images/caching-workflow.png)
+
```
dalmia, line 13: can you make a list of requests that we are caching? 

GET plio
GET user

docs/CACHING.md
rathorevaibhav: Okay, makes sense. I'll update it or maybe create a separate one to explain the flow better.

plio/models.py
rathorevaibhav: Right! I'll update.

plio/tests.py
```diff
@@ -464,6 +475,46 @@ def test_items_sorted_with_time(self):
         self.assertEqual(response.data["items"][0]["id"], item_2.id)
         self.assertEqual(response.data["items"][1]["id"], item_1.id)
 
+    def test_retrieving_video_sets_instance_cache(self):
```
dalmia, line 478: why is the name `retrieving_video` - should be `retrieving_plio`.

plio/cache.py
rathorevaibhav: Yes, questions and items I'll add.

We are currently sending plio details in the session response (along with a lot of other information). Since we are not making any GET requests for the session, so unlike Plio response, I don't think the whole session response will be cached. So we don't need session cache key here. I'll remove.

But we can do it in this way. Session response includes:
1. Plio (can come from cache)
2. User (can come from cache)
3. Experiment (can come from cache)
4. Last Event (won't be cached)
5. Session Answers (probably won't be cached)

plio/cache.py
rathorevaibhav: This doesn't. It simply returns an integer for how many keys got deleted. 0 or more.

plio/cache.py
rathorevaibhav: It won't. It simply returns True or False.

users/models.py
rathorevaibhav: This didn't work and started throwing duplicate parameter `_`. Can you clarify why you suggested this change?

docs/CACHING.md
```diff
@@ -9,6 +9,7 @@ Plio's caching mechanism can be explained in three simple steps:
 2. Searches for the requested data in the cache
 3. If not present, query the database and save the response in cache for future requests
 
+##### Caching workflow explained via flow diagram
 ![Overview of caching](images/caching-workflow.png)
```
dalmia, line 13: Here, the part about cache invalidation is still missing in the diagram.

docs/CACHING.md
```diff
@@ -17,7 +18,18 @@ The calculation of the cache keys are based on the model instances. For example,
...
 
-The new cache will be set during the first fresh response calculated from the database and will be used for subsequent requests.
+The new cache will be set when the first fresh response is calculated from the database and will be used for subsequent requests.
+
+##### Cache invalidation explained via flow diagram
+![Overview of caching](images/cache-invalidation-workflow.png)
+
+
+### Current cached data
+We only have implemented caching for data that has high GET requests. The following resources have been cached:
```
dalmia, line 31:
```suggestion
We have only implemented caching for models with a high number of GET requests. The following resources have been cached:
```

plio/cache.py
```diff
@@ -1,15 +1,27 @@
 from django.core.cache import cache
+from django.db import connection
 
 
 def get_cache_key(instance):
-    """Calculates cache key for an instance"""
+    """Calculates the cache key for an instance based on selected schema and """
```
dalmia, line 6: the statement seems incomplete.

users/tests.py
```diff
@@ -1,12 +1,14 @@
 import json
 from rest_framework import status
 from django.urls import reverse
+from django.core.cache import cache
 
 from oauth2_provider.models import AccessToken, RefreshToken
 from users.models import User
 from users.models import OneTimePassword, OrganizationUser
 from plio.tests import BaseTestCase
 from organizations.models import Organization
```
dalmia, line 10:
```suggestion
from organizations.models import Organization
from plio.tests import BaseTestCase
```

plio/models.py
```diff
@@ -158,3 +158,17 @@ def video_update_cache(sender, instance, **kwargs):
...
+def item_update_cache(sender, instance, **kwargs):
+    # invalidate saved cache for the plio
+    invalidate_cache_for_instance(instance.plio)
+
+
+@receiver(post_save, sender=Question)
+@receiver(post_delete, sender=Question)
+def question_update_cache(sender, instance, **kwargs):
+    # invalidate saved cache for the plio
+    invalidate_cache_for_instance(instance.item.plio)
```
dalmia, line 174: here, you'll have to invalidate the cache for `instance.item` too right?

organizations/tests.py
```diff
@@ -1,16 +1,26 @@
...
 from rest_framework import status
 from django.urls import reverse
 
+from django.core.cache import cache
+
+from plio.tests import BaseTestCase
+from organizations.models import Organization
+
+from plio.cache import get_cache_key
+from users.models import OrganizationUser
```
dalmia, line 10:
```suggestion
from plio.tests import BaseTestCase
from plio.cache import get_cache_key
from organizations.models import Organization
from users.models import OrganizationUser
```

organizations/tests.py
```diff
@@ -39,3 +49,57 @@ def test_superuser_can_list_organizations(self):
...
+            organization=self.organization_2, user=self.user, role=self.org_admin_role
+        )
+
+        # check cache after updating organization user association
+        # this time the user should have an organization associated
+        self.assertEqual(len(cache.get(cache_key_name)["organizations"]), 1)
+        self.assertEqual(
+            cache.get(cache_key_name)["organizations"][0]["name"],
+            self.organization_2.name,
+        )
```
dalmia, line 87: shouldn't this be a part of `OrganizationUserTestCase`?

organizations/tests.py
```diff
@@ -39,3 +49,57 @@ def test_superuser_can_list_organizations(self):
...
+        OrganizationUser.objects.create(
+            organization=self.organization_2, user=self.user, role=self.org_admin_role
+        )
+
+        # check cache after updating organization user association
+        # this time the user should have an organization associated
+        self.assertEqual(len(cache.get(cache_key_name)["organizations"]), 1)
+        self.assertEqual(
+            cache.get(cache_key_name)["organizations"][0]["name"],
+            self.organization_2.name,
```
dalmia, line 86: note for self: look at the workflow diagram for invalidation to understand this better.
essentially, after the cache invalidation takes place, the request fetches the data again from the DB in order to send the response. In that case, this new data is stored in the cache.

organizations/tests.py
```diff
@@ -39,3 +49,57 @@ def test_superuser_can_list_organizations(self):
...
+        # user cache should be deleted after organization update
+        self.assertEqual(len(cache.keys(cache_key_name)), 0)
+
+        # request user again so that we can check if the cache is updated
+        self.client.get(reverse("users-detail", kwargs={"pk": self.user.id}))
+
+        # verify cache data has now the updated value
+        self.assertEqual(
+            cache.get(cache_key_name)["organizations"][0]["name"], org_new_name
+        )
```
dalmia, line 105: one point of distinction which I feel needs to be made:
for the request above, there was a PUT request to OrganisationUser and hence, the retrieval from DB post cache invalidation happened in the same request as the request had to return the new OrganisationUser object in the response. 

However, here, for a PATCH request, once the cache is invalidated, the same request won't fetch the updated object from DB and hence, we need to make a separate GET request to fill the cache. 

This also means that the workflow diagram shouldn't be mentioning PUT/PATCH/DELETE in the types of requests for which the diagram is valid right? We should have a separate cache invalidation diagram for PATCH.

plio/models.py
```diff
@@ -158,3 +158,17 @@ def video_update_cache(sender, instance, **kwargs):
     plios = Plio.objects.filter(video_id=instance.id)
     # invalidate saved cache for the plios
     invalidate_cache_for_instances(plios)
+
+
+@receiver(post_save, sender=Item)
+@receiver(post_delete, sender=Item)
+def item_update_cache(sender, instance, **kwargs):
+    # invalidate saved cache for the plio
+    invalidate_cache_for_instance(instance.plio)
```
dalmia, line 167: this should also invalidate the cache for question linked to that item too right?

plio/tests.py
```diff
@@ -774,6 +807,42 @@ def test_deleting_plio_deletes_items(self):
         response = self.client.get(f"/api/v1/items/{self.item.id}/")
         self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
 
+    def test_updating_item_updates_linked_plio_instance_cache(self):
```
dalmia, line 810: need to also test that updating item updates the question that is linked to the item as the item details are also returned in the question Serializer.

plio/models.py
```diff
@@ -158,3 +158,17 @@ def video_update_cache(sender, instance, **kwargs):
     plios = Plio.objects.filter(video_id=instance.id)
     # invalidate saved cache for the plios
     invalidate_cache_for_instances(plios)
+
```
dalmia, line 161: we have to be careful of which objects have been included in the response of which other objects through `to_representation` in the serializer.

organizations/tests.py
rathorevaibhav: I think you are right. There's something here that's not clearly mentioned in the diagram.

When we're making a PUT/PATCH/DELETE request for the same resource (like user in this test case), it updates the cache with these steps:
1. post_save signal deletes the cache
2. the serializer for that resource creates the cache with new values

But when we're making a PUT/PATCH/DELETE request for a related resource and not the same (like organizations for user in this case), it only deletes the user cache and not creates it again as UserSerializer didn't run in organization update/delete request.

So the diagram is partially correct. I'll see how I can fit it in the diagram.

plio/models.py
rathorevaibhav: not really as we're not caching item. Only Plio responses which have items included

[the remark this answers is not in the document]
rathorevaibhav: Image is not updated. A new image entry is created and question instance is updated with the image id.

plio/models.py
rathorevaibhav: so we're not caching question also. Only the whole plio response. So we are resetting the whole plio cache and not individual question.

plio/models.py
deepansh96: And we're not caching item or question because we don't use any GET request for items or questions instances right?

organizations/tests.py
```diff
@@ -39,3 +48,57 @@ def test_superuser_can_list_organizations(self):
...
+        superadmin_user = User.objects.create(mobile="+919988776655", is_superuser=True)
+        superadmin_access_token = get_new_access_token(
+            superadmin_user, self.application
+        )
+        superadmin_client.credentials(
+            HTTP_AUTHORIZATION="Bearer " + superadmin_access_token.token
+        )
+
+        # verify cache data doesn't exist by default
+        cache_key_name = get_cache_key(self.user)
```
deepansh96, line 67: I think I didn't understand how does this line is verifying that cache data doesn't exist

deepansh96: There should be this line here as well, right?
`self.assertEqual(len(cache.keys(cache_key_name)), 0)`

plio/cache.py
```diff
@@ -3,7 +3,7 @@
 
 
 def get_cache_key(instance):
-    """Calculates the cache key for an instance based on selected schema and """
+    """Calculates the cache key for an instance based on selected tenant schema"""
```
dalmia, line 6: since the tenant is take from the connection and not passed as a param
```suggestion
    """Calculates the cache key for an instance based on the current tenant schema"""
```

plio/models.py
dalmia: ah sorry - makes sense!

organizations/tests.py
dalmia: I think you can leave out the assert part here - and leave everything else. And use the version with assert under `OrganizationUserTestCase`. 

Also, I don't seem to be able to understand why this is working - as in, without a GET request first, why is the cache for the user getting updated? Upon updating organization user, the cache gets invalidated. But where is it getting filled up again with the updated value?

organizations/tests.py
dalmia: Yep, that would be helpful. Although I don't seem to see any changes to the image. is that intentional? to convey this just in the markdown files?

plio/tests.py
```diff
@@ -932,6 +1001,49 @@ def test_deleting_plio_deletes_questions(self):
...
+        self.assertEqual(len(cache.keys(cache_key_name)), 0)
+
+        # request plio again via API to generate cache
+        self.client.get(reverse("plios-detail", kwargs={"uuid": plio.uuid}))
+
+        # plio cache should exist now
+        self.assertEqual(len(cache.keys(cache_key_name)), 1)
+        self.assertEqual(
+            cache.get(cache_key_name)["items"][0]["details"]["text"], question_text
+        )
```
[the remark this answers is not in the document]
dalmia: Got it. let's keep it then.

organizations/tests.py
dalmia: as you mentioned in the docs as well, the user cache should be deleted but not updated right?
Screenshot 2021-09-24 at 3 52 05 PM
he

plio/models.py
rathorevaibhav: @deepansh96 Yes, for now. Though we do have endpoints but I don't think we're using GET endpoints for item and question. In case we do, we can apply caching there as well.

organizations/tests.py
rathorevaibhav: Yep. I am a bit not sure exactly how to represent it diagrammatically. It felt like the explanation is text-heavy.

organizations/tests.py
rathorevaibhav: @dalmia I think your full comment didn't come through. can you share again?

Yes, in case related instances are updated, the primary instance cache gets deleted and not re-created. The cache for an instance is created in the corresponding serializer. Since updating related instance doesn't call the primary instance serializer, hence the update cache doesn't happen. Only deletion.

organizations/tests.py
rathorevaibhav: Yes. Added. Thanks for catching.

organizations/tests.py
rathorevaibhav: hey sorry I saw your previous comment just now.

What you wrote makes sense and that's my understanding as well. I'm trying to figure out why it's working (updating automatically). That's weird. Will get back to you.

docs/CACHING.md
```diff
@@ -30,20 +30,21 @@ The new cache will be set when the first fresh response is calculated from the d
...
 
 2. #### User
-    - User cache is created (if not exists) when there is a GET request for a user instance.
-    - User cache is created or updated when there is a PUT/PATCH/DELETE request for a user instance.
-    - User cache is deleted when there is a PUT/PATCH/DELETE request for any of the following related instances for a user instance:
-        - OrganizationUser
+    - User cache is created when there is a retrieve request for a user instance.
+    - User cache is re-created when there is a create, update or delete request for any of the following related instances:
+      - User
+      - OrganizationUser (as the UserSerializer is called when modifying OrganizationUser)
```
dalmia, line 46:
```suggestion
      - OrganizationUser (as the `UserSerializer` is called when `OrganizationUser` is modified)
```

PR #259, opened 2021-11-29
Feature/cypress ci

rathorevaibhav:
## Summary

1. CI support for cypress in backend repo
2. Blank ENV variables values replaced by empty strings to avoid errors in new Docker Desktop docker-compose runs
3. New command `pliooauth2application` that creates a default application and client credentials for API programmatically. Required to let Cypress make API calls from frontend to backend.
4. The OAuth2 application only runs if `DEFAULT_OAUTH2_CLIENT_SETUP` env variable is set to True. If not provided, the command considers it false.

## Test Plan

- [ ] Wrote tests
- [ ] Tested locally
- [ ] Tested on staging
- [ ] Tested on production


.env.example
```diff
@@ -1,76 +1,81 @@
...
 SMS_DRIVER='sns'
 
 # Sentry Data Source Name
-SENTRY_DSN=
+SENTRY_DSN=''
+
+# Default OAuth2 API Client
+DEFAULT_OAUTH2_CLIENT_SETUP=True
+DEFAULT_OAUTH2_CLIENT_ID=''
+DEFAULT_OAUTH2_CLIENT_SECRET=''
```
dalmia, line 81: the entry for these new env variables are missing in `docs/ENV.md` right? Even if no one needs to set them up, would be good to explicitly mention that or else someone new might be confused.

.env.example
```diff
@@ -1,76 +1,81 @@
...
-DEFAULT_FROM_EMAIL=Plio <no-reply@plio.in>
-EMAIL_HOST_USER= # sender's email-id
+DEFAULT_FROM_EMAIL='Plio <no-reply@plio.in>'
+EMAIL_HOST_USER='' # sender's email-id
 # password associated with above email-id
 # for accounts with 2-factor authentication, you have to create a application-specific
 # passwords. More on that here:
 # https://dev.to/abderrahmanemustapha/how-to-send-email-with-django-and-gmail-in-production-the-right-way-24ab
-EMAIL_HOST_PASSWORD=
+EMAIL_HOST_PASSWORD=''
```
dalmia, line 68: we are not using these email settings anymore. Can you please remove them from the codebase and the environment files?

entrypoint.sh
```diff
@@ -16,6 +16,9 @@ python manage.py pliotenant
 # create default superuser
 python manage.py pliosuperuser
 
+# create default OAuth2 API client credentials
+python manage.py pliooauth2application
```
dalmia, line 20: we can avoid using `plio` in this file, right? seems redundant. not only for this command for the command above (`pliosuperuser` too)? might be better to use something that is more meaningful like `createsuperuser` or `createoauth2application`

plio/management/commands/pliooauth2application.py
```diff
@@ -0,0 +1,32 @@
+from django.core.management.base import BaseCommand
+import os
+from oauth2_provider.models import Application
+
+
+class Command(BaseCommand):
+    help = "Creates OAuth2 client credentials"
```
dalmia, line 7: Is this accurate?
```suggestion
    help = "Creates a Default OAuth2 client credentials for CI"
```

plio/management/commands/pliooauth2application.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def handle(self, *args, **options):
+        default_oauth2_client_setup = os.environ.get(
+            "DEFAULT_OAUTH2_CLIENT_SETUP", False
+        )
+        default_oauth2_client_id = os.environ.get("DEFAULT_OAUTH2_CLIENT_ID")
+        default_oauth2_client_secret = os.environ.get("DEFAULT_OAUTH2_CLIENT_SECRET")
+        if default_oauth2_client_setup:
```
dalmia, line 15: we can also rewrite it as:
```python
if not default_oauth2_client_setup:
    print("Default OAuth2 client setup is disabled. Skipping.")
    return
.
.
.
```
and so on for the other `if-else` blocks, right?

This will avoid a lot of branching in the code and early exit - also, helps with readability IMO.

plio/management/commands/pliooauth2application.py
```diff
@@ -0,0 +1,32 @@
...
+
+class Command(BaseCommand):
+    help = "Creates OAuth2 client credentials"
+
+    def handle(self, *args, **options):
+        default_oauth2_client_setup = os.environ.get(
+            "DEFAULT_OAUTH2_CLIENT_SETUP", False
+        )
+        default_oauth2_client_id = os.environ.get("DEFAULT_OAUTH2_CLIENT_ID")
+        default_oauth2_client_secret = os.environ.get("DEFAULT_OAUTH2_CLIENT_SECRET")
```
dalmia, line 14: plus, these values won't have to be extracted if `default_oauth2_client_setup` is false, right?

So, we'll have to set these variables while deploying on staging too, right? need to ensure that we don't forget it - so, better to set these variables before merging itself.

dalmia (changes requested):

.env.example
rathorevaibhav: oh no. I completely missed. Yes, will add.

entrypoint.sh
rathorevaibhav: Sure.

plio/management/commands/pliooauth2application.py
rathorevaibhav: makes sense. yes. will do.

.env.example
rathorevaibhav: Okay. I've removed all the email settings. They were only present in `.env.example` and `settings.py`.

.env.example
rathorevaibhav: Done.

rathorevaibhav:
@dalmia @deepansh96 any ideas on the code coverage reduction? All I did was remove some configs from settings.py where it's showing the reduction in code coverage

dalmia:
> @dalmia @deepansh96 any ideas on the code coverage reduction? All I did was remove some configs from settings.py where it's showing the reduction in code coverage

I'd say not to worry about it.

plio/management/commands/createoauth2application.py
```diff
@@ -0,0 +1,43 @@
...
+        )
+        if not default_oauth2_client_setup:
+            print("Default OAuth2 client setup is disabled. Skipping.")
+            return
+
+        default_oauth2_client_id = os.environ.get("DEFAULT_OAUTH2_CLIENT_ID")
+        default_oauth2_client_secret = os.environ.get("DEFAULT_OAUTH2_CLIENT_SECRET")
+
+        if not default_oauth2_client_id or not default_oauth2_client_secret:
+            print("No default OAuth2 client id and secret provided. Skipping.")
```
dalmia, line 21:
```suggestion
            print("No default OAuth2 client id or secret provided. Skipping.")
```

plio/management/commands/createoauth2application.py
```diff
@@ -0,0 +1,43 @@
...
+
+        print(
+            "Default OAuth2 client id and secret provided. Creating default application."
+        )
+        application = Application.objects.filter(
+            client_id=default_oauth2_client_id
+        ).first()
+
+        if application:
+            print("Default OAuth2 client id already exists. Skipping.")
```
dalmia, line 32:
```suggestion
            print("An application with the default OAuth2 client ID already exists. Skipping.")
```

docs/ENV.md
```diff
@@ -159,3 +159,16 @@ Plio supports error monitoring on your app with [Sentry](https://sentry.io/). Vi
 
 #### `SENTRY_DSN`
 The Sentry Data Source Name. This is a unique link that identifies your Sentry project and sends error reports to Sentry dashboard. You can get the value of DSN for your projects from `Sentry dashboard > Settings > Project > {PROJECT_NAME} > SDK Setup > Client Keys (DSN)`.
+
+
+### OAuth2 Client Credentials
+Using the `python manage.py createoauth2application` command, you can create an OAuth2 client application for your app. For that, the following environment variables are needed.
+
+#### `DEFAULT_OAUTH2_CLIENT_SETUP`
+Boolean value. Defaults to `False` if not set. The command `createoauth2application` will work only if this is set to True.
```
deepansh96, line 168:
```suggestion
Boolean value. Defaults to `False` if not set. The command `createoauth2application` will work only if this is set to `True`.
```

plio/management/commands/pliooauth2application.py
deepansh96: It is optional but it is present in the `entrypoint.sh` so it will get run anyhow right?
But that shouldn't be an issue for the cases where we already have a client id and secret information in the .env file, in that case, it'll just skip and won't create. So no need of any change in local/staging/prod

This is specifically so that the cypress can set up a client when setting up the repo for first time.
Am i correct?

rathorevaibhav:
> It is optional but it is present in the entrypoint.sh so it will get run anyhow right?
But that shouldn't be an issue for the cases where we already have a client id and secret information in the .env file, in that case, it'll just skip and won't create. So no need of any change in local/staging/prod
This is specifically so that the cypress can set up a client when setting up the repo for first time.
Am i correct?

Yes, that's right.

PR #296, opened 2022-01-14
Adds support for deleting items in bulks + deleting linked image if question is deleted

### LINKED ISSUE #294: delete the linked image if a question is deleted

### LINKED ISSUE #295: Support for bulk deletion of items

dalmia:
Fixes #295 
Fixes #294 

## Test Plan

- [x] Wrote tests
- [x] Tested locally
- [x] Tested on staging
- [ ] Tested on production


plio/signals.py
```diff
@@ -28,3 +28,11 @@ def item_update_cache(sender, instance, **kwargs):
 def question_update_cache(sender, instance, **kwargs):
     # invalidate saved cache for the plio
     invalidate_cache_for_instance(instance.item.plio)
+
+
+@receiver([post_save], sender=Question)
```
deepansh96, line 33: Why not post_delete?

plio/signals.py
```diff
@@ -28,3 +28,11 @@ def item_update_cache(sender, instance, **kwargs):
...
+def question_delete_image_on_deletion(sender, instance, **kwargs):
```
deepansh96, line 34:
```suggestion
def delete_linked_image_on_question_deletion(sender, instance, **kwargs):
```
How about this?

deepansh96, line 34: Can you also add one liner of docstring? Because in the code, that part is confusing
`if instance.deleted is not None` -- I mean we're doing this when the question instance is deleted right? and the code says `if instance.deleted is not None`

plio/views.py
```diff
@@ -558,6 +558,34 @@ def duplicate(self, request, pk):
         item.save()
         return Response(self.get_serializer(item).data)
 
+    @action(methods=["delete"], detail=False)
```
deepansh96, line 561: Add permissions?
item bulk deletion should be a protected action :p

plio/views.py
```diff
@@ -558,6 +558,34 @@ def duplicate(self, request, pk):
...
+    def bulk_delete(self, request):
+        """deletes items whose ids have been provided"""
+        if "id" not in request.data:
+            return Response(
+                {"detail": "id not provided"},
```
deepansh96, line 566:
```suggestion
                {"detail": "item id(s) not provided"},
```

plio/views.py
```diff
@@ -558,6 +558,34 @@ def duplicate(self, request, pk):
...
+
+        items_to_delete = Item.objects.filter(pk__in=ids_to_delete)
+        if len(items_to_delete) != len(ids_to_delete):
+            return Response(
+                {"detail": "one or more of the ids provided do not exist"},
+                status=status.HTTP_404_NOT_FOUND,
+            )
+
+        items_to_delete.delete()
+        return Response("deletion successful")
```
deepansh96, line 587: Also add an appropriate status code here?

plio/signals.py
```diff
@@ -28,3 +28,11 @@ def item_update_cache(sender, instance, **kwargs):
...
+    # check if any image is linked to the instance
+    if instance.deleted is not None and instance.image is not None:
+        # delete that image as well
+        instance.image.delete()
```
deepansh96, line 38: is this soft delete or what? Will the entry get deleted from the DB as well?
If yes, then we should also delete the image on S3 right? Otherwise we'll end up with a lot of images on S3 folder which are not connected to anything. I don't think it's much of a problem but things will be neat and tidy, no loose ends

plio/views.py
```diff
@@ -558,6 +558,34 @@ def duplicate(self, request, pk):
...
+            )
+
+        items_to_delete = Item.objects.filter(pk__in=ids_to_delete)
+        if len(items_to_delete) != len(ids_to_delete):
+            return Response(
+                {"detail": "one or more of the ids provided do not exist"},
+                status=status.HTTP_404_NOT_FOUND,
+            )
+
+        items_to_delete.delete()
```
deepansh96, line 586: this is bulk delete?
Just `delete()`? :p

deepansh96: And this is soft delete right?
Should we just permanently delete the items?

deepansh96 (changes requested):

plio/signals.py
dalmia: haan I thought so at first too. turns out that because we are doing soft_delete, `post_delete` is never invoked only.

plio/views.py
dalmia: I was a little confused ki does adding a custom action inherit the permissions of the viewset or not? if it does, we won't need to add. will research

plio/views.py
dalmia: clarified that default is 200!

plio/signals.py
dalmia: soft_delete mention

plio/views.py
dalmia: check this actually

plio/signals.py
dalmia: mention soft delete

plio/views.py
dalmia: the action inherits the permissions of the viewsets. the option given here is in the case if we want to override it.

plio/signals.py
```diff
@@ -31,8 +31,12 @@ def question_update_cache(sender, instance, **kwargs):
 
 
 @receiver([post_save], sender=Question)
-def question_delete_image_on_deletion(sender, instance, **kwargs):
-    # check if any image is linked to the instance
+def delete_linked_image_on_question_deletion(sender, instance, **kwargs):
+    # since we are using soft deletion, the `post_delete` signal is never called
+    # also, the way deletion works under the hood when using soft delete is that
+    # just sets the `deleted` attribute and updates the instance. so, by default,
+    # the deleted attribute is None. When we execute soft_delete, that attribute gets set
```
deepansh96, line 38: Some inconsistency
`so, be default...` here, a new sentence starts with a lower case letter
`When we execute ... ` here, a new sentence starts with an upper case letter

We should decide on a format or we should ignore such things and not include them in the review.

Mentioning this because in another of backend PR of mine, the review included comments starting with caps and that they should be turned to lower case

I don't mind too much about the upper/lower casing of comments so let me know whatever you decide

plio/signals.py
```diff
@@ -28,3 +28,15 @@ def item_update_cache(sender, instance, **kwargs):
...
+
+
+@receiver([post_save], sender=Question)
+def delete_linked_image_on_question_deletion(sender, instance, **kwargs):
+    # since we are using soft deletion, the `post_delete` signal is never called
+    # also, the way deletion works under the hood when using soft delete is that
+    # just sets the `deleted` attribute and updates the instance. so, by default,
+    # the deleted attribute is None. When we execute soft_delete, that attribute gets set
+    if instance.deleted is not None and instance.image is not None:
+        # if any image is linked to the question instance
```
deepansh96, line 40:
```suggestion
        # if any image is linked to the question instance,
```

plio/views.py
deepansh96: can you also do these quick changes in other viewsets as well? shouldn't take long

PR #304, opened 2022-01-19
Checks whether correct params have been passed during OTP verification

### LINKED ISSUE #301: For otp verification, check that both mobile and OTP have been given

Manyakalra3:
Fixes #301

## Summary
Tests are written for testing whether Params (mobile and otp) have been passed during the OTP verification.

## Test Plan

- [x] Wrote tests
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production
- [ ] <strike>If adding any new environment variable:<strike>
    - [ ] update `docs/ENV.md`

### DIFF
--- a/users/tests.py
+++ b/users/tests.py
@@ -28,14 +28,14 @@ class OtpAuthTestCase(BaseTestCase):
         otp_exists = OneTimePassword.objects.filter(mobile=self.user.mobile).exists()
         self.assertTrue(otp_exists)
 
-  
     def test_otp_not_passed(self):
-        #check if otp is present
-        response = self.client.post(reverse("verify-otp"),{"mobile": self.user.mobile})
+        # check if otp is present
+        response = self.client.post(reverse("verify-otp"), {"mobile": self.user.mobile})
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
         self.assertEqual(response.data["detail"], "OTP Not Provided")
+
     def test_mobile_not_passed(self):
-        #check whether mobile is defined
+        # check whether mobile is defined
         response = self.client.post(reverse("verify-otp"))
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
         self.assertEqual(response.data["detail"], "mobile not defined")
--- a/users/tests.py
+++ b/users/tests.py
@@ -28,6 +28,18 @@ class OtpAuthTestCase(BaseTestCase):
         otp_exists = OneTimePassword.objects.filter(mobile=self.user.mobile).exists()
         self.assertTrue(otp_exists)
 
+  
+    def test_otp_not_passed(self):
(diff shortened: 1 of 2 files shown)


users/views.py
```diff
@@ -180,6 +180,16 @@ def request_otp(request):
...
+    if "mobile" not in request.data:
+        return Response(
+            {"detail": "mobile not defined"},
+            status=status.HTTP_400_BAD_REQUEST,
+        )
+    if "otp" not in request.data:
+        return Response(
+            {"detail": "OTP Not Provided"},
+            status=status.HTTP_400_BAD_REQUEST,
+        )
```
dalmia, line 192: you could club this together:
```suggestion
    for key in ['mobile', 'otp']:
        if key not in request.data:
            return Response(
                {"detail": f"{key} not provided"},
                status=status.HTTP_400_BAD_REQUEST,
            )
```

users/tests.py
```diff
@@ -28,6 +28,18 @@ def test_guest_can_request_for_otp(self):
         otp_exists = OneTimePassword.objects.filter(mobile=self.user.mobile).exists()
         self.assertTrue(otp_exists)
 
+    def test_otp_not_passed(self):
```
dalmia, line 31: so, the name of the test should specify what should happen given a particular set of inputs. So, in this case, a better name could be: `test_verifying_otp_fails_when_otp_not_passed`

users/tests.py
```diff
@@ -28,6 +28,18 @@ def test_guest_can_request_for_otp(self):
...
+        # check if otp is present
+        response = self.client.post(reverse("verify-otp"), {"mobile": self.user.mobile})
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.data["detail"], "OTP Not Provided")
+
+    def test_mobile_not_passed(self):
```
dalmia, line 37: similar comment as above!

users/tests.py
```diff
@@ -28,6 +28,18 @@ def test_guest_can_request_for_otp(self):
         otp_exists = OneTimePassword.objects.filter(mobile=self.user.mobile).exists()
         self.assertTrue(otp_exists)
 
+    def test_otp_not_passed(self):
+        # check if otp is present
+        response = self.client.post(reverse("verify-otp"), {"mobile": self.user.mobile})
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.data["detail"], "OTP Not Provided")
```
dalmia, line 35: your error messages are different. In one case, it is "not provided" and in the other case, it is "not defined". Given that the issue is the same (some parameter not being provided), the error messages should be consistent too!

dalmia (changes requested):

dalmia (approved):

PR #306, opened 2022-01-19
checks whether correct params for mobile have been passed during OTP request

### LINKED ISSUE #300: Under OTP request, add check for the case when mobile has not been given

gurleen3:
Fixes #300

## Summary
Checks whether correct params have been passed during OTP request 

Tests are written for testing whether the correct params(mobile) have been passed while requesting otp. 

## Test Plan

- [x] Wrote tests
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production
- [ ] ~If adding any new environment variable:~
    - [ ] update `docs/ENV.md`

### DIFF
--- a/users/tests.py
+++ b/users/tests.py
@@ -11,17 +11,16 @@ from plio.tests import BaseTestCase
 from plio.cache import get_cache_key
 
 
-class OtpAuthTestCase(BaseTestCase):  
+class OtpAuthTestCase(BaseTestCase):
     """Tests the OTP functionality."""
 
     def setUp(self):
         super().setUp()
         # unset client credentials token so that the subsequent API calls goes as guest
         self.client.credentials()
+
     def test_request_otp_mobile_not_passed(self):
-        response = self.client.post(
-            reverse("request-otp")
-        )
+        response = self.client.post(reverse("request-otp"))
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
 
     def test_guest_can_request_for_otp(self):
--- a/users/tests.py
+++ b/users/tests.py
@@ -22,7 +22,7 @@ class OtpAuthTestCase(BaseTestCase):
     def test_requesting_otp_fails_when_mobile_not_passed(self):
         response = self.client.post(reverse("request-otp"))
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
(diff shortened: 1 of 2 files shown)


users/tests.py
```diff
@@ -11,13 +11,18 @@
...
 
-class OtpAuthTestCase(BaseTestCase):
+class OtpAuthTestCase(BaseTestCase):  
     """Tests the OTP functionality."""
 
     def setUp(self):
         super().setUp()
         # unset client credentials token so that the subsequent API calls goes as guest
         self.client.credentials()
+    def test_request_otp_mobile_not_passed(self):
```
dalmia, line 21: all functions within classes should be separate by a space. Currently, there is no space before this function.

dalmia, line 21: leaving a few similar feedback points as I mentioned on Manya's PR: the name of the test should specify what should happen given a particular set of inputs. So, in this case, a better name could be: `test_requesting_otp_fails_when_mobile_not_passed`

users/tests.py
```diff
@@ -11,13 +11,18 @@
...
+        response = self.client.post(
+            reverse("request-otp")
+        )
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
dalmia, line 25: do check for the error message as well - the way Manya has done in her PR: https://github.com/avantifellows/plio-backend/pull/304

dalmia (changes requested):

users/tests.py
```diff
@@ -11,13 +11,18 @@
 from plio.cache import get_cache_key
 
 
-class OtpAuthTestCase(BaseTestCase):
+class OtpAuthTestCase(BaseTestCase):
```
deepansh96, line 14: There's an extra space character added here
Usually it's best to see the "Files Changed" tab in a pull request yourself so you can see what all got changed from your commits

deepansh96: I guess pre commit is also failing. All the spacing issues should be fixed by pre commit

gurleen3: done

users/tests.py
gurleen3: corrected

users/tests.py
gurleen3: corrected it

PR #310, opened 2022-01-21
Removes support for plio-analytics and BIGQUERY

### LINKED ISSUE #309: Remove support for plio-analytics

dalmia:
Fixes #309 

## Summary
- removed the `list_uuid` endpoint and instead updated the `list` endpoint to return a paginated list of plio details with each plio object also containing the items associated with that plio (this is because of the way duplicate is implemented - created an issue to fix that later) + number of unique views.
- removed support for BigQuery and making all queries directly on RDS
- added a new endpoint for computing metrics for a plio
- removed everything related to BigQuery and plio-analytics

## Test Plan

- [x] Wrote tests
- [x] Tested locally
- [x] Tested on staging
- [ ] Tested on production
- [x] If adding or removing any environment variable:
    - [ ] update `docs/ENV.md`


plio/queries.py
```diff
@@ -38,11 +35,11 @@ def get_sessions_dump_query(plio_uuid: str, schema: str, mask_user_id: bool = Tr
             CASE
                 WHEN {str(mask_user_id).lower()} THEN COALESCE(users.email, users.mobile, CONCAT('unique_id:', users.unique_id))
             ELSE
-                {'TO_HEX(MD5(CAST(session.user_id as STRING)))' if BIGQUERY['enabled'] else 'MD5(session.user_id::varchar(255))'}
+                'MD5(session.user_id::varchar(255))'
```
dalmia, line 38: TODO: fix this

plio/views.py
```diff
@@ -183,17 +177,38 @@ def list_uuid(self, request):
...
+            }
+
+            # add the number of unique viewers and items corresponding
+            # to the plio in each plio object
+            for index, plio in enumerate(page):
+                page[index]["num_views"] = plio_id_to_count.get(plio["id"], 0)
+                page[index]["items"] = ItemSerializer(
+                    queryset[index].item_set, many=True
+                ).data
+
```
deepansh96, line 204: A couple of things here
- You're calculating unique viewers for each plio here and saving that in a temporary dictionary and attaching that information in the response, right? In a file called `ordering.py`, there is already a piece of code jisme I calculated unique viewers for each plio and save that information in the queryset itself. I think it would be best to reuse code in some way, as it does the same thing. 
- Also, when I do just `list`, this piece of code runs and I get number of unique views for each of the 5 plios. But let's suppose after that I want to order by views, so I click the views column in the table in frontend. That makes a call to `list` again but this time with the ordering filter. So what happens is that the compiler goes to `ordering.py`, calculates the unique views for each plio, sorts them according to that, and then AGAIN finds the num unique viewers for each plio (the code that you wrote). So double work is being done here. After the compiler gets through `ordering.py`, the information of num unique viewers is already stored in the queryset so you can reuse that.

plio/views.py
```diff
@@ -183,17 +177,38 @@ def list_uuid(self, request):
...
+                for unique_user_count in unique_user_counts
+            }
+
+            # add the number of unique viewers and items corresponding
+            # to the plio in each plio object
+            for index, plio in enumerate(page):
+                page[index]["num_views"] = plio_id_to_count.get(plio["id"], 0)
+                page[index]["items"] = ItemSerializer(
+                    queryset[index].item_set, many=True
+                ).data
```
deepansh96, line 203: Why do we need items when i'm just doing `list` in the homepage? I just need top level plio details right?

plio/views.py
```diff
@@ -335,6 +350,203 @@ def copy(self, request, uuid):
...
+    )
+    def metrics(self, request, uuid):
+        """Returns usage metrics for the plio"""
+        # return 404 if user cannot access the object
+        # else fetch the object
+        plio = self.get_object()
+
+        # no sessions have been created for the plio: return
+        if not Session.objects.filter(plio=plio.id):
+            return Response({})
```
deepansh96, line 366: how is this being handled at the frontend?

plio/views.py
```diff
@@ -335,6 +350,203 @@ def copy(self, request, uuid):
...
+        num_views = len(df)
+        average_watch_time = df["watch_time"].mean()
+
+        # retention at one minute
+        if plio.video.duration is None or plio.video.duration < 60:
+            # the metric is not applicable in this case
+            percent_one_minute_retention = None
+        else:
+            # convert "0,1,0" to ["0", "1", "0"]
+            df["retention"] = df["retention"].apply(lambda row: row.split(","))
```
deepansh96, line 402: Why not move the metrics code to some other file? Will help us as well if we are going to add more metrics, will also keep the `plio/views.py` cleaner

Like there's a file for all the queries, `queries.py` so it makes sense to add these queries there as well

deepansh96 (changes requested):

deepansh96:
- You'll also remove all the instances related to the analytics service right? Auth0, cognito, ECR instances etc for both staging + production?
- I had this stuff configured in my .env file. These instances will also be shut down and deleted right?
image

PR #312, opened 2022-01-25
Increased expiry time of access token

### LINKED ISSUE #311: Increase expiry time of access token from 1 day to 7 days
Currently, an access token gets expired in 1 day.
This leads to 401 calls and re-authentication calls for all active users every day.

Increasing the expiry time to 7 days will avoid so many 401 / re-authentication calls, as well as the access_token/refresh_token table in the DB will not get filled so fast

deepansh96:
Fixes #311 

## Test Plan

- [ ] ~Wrote tests~
- [x] Tested locally
- [x] Tested on staging
- [ ] Tested on production
    - [ ] update `docs/ENV.md`


plio/settings.py
```diff
@@ -264,7 +264,7 @@
 API_APPLICATION_NAME = "plio"
 
 OAUTH2_PROVIDER = {
-    "ACCESS_TOKEN_EXPIRE_SECONDS": 60 * 60 * 24,  # 1 day
+    "ACCESS_TOKEN_EXPIRE_SECONDS": 60 * 60 * 24 * 7,  # 7 days
```
dalmia, line 267: after checking the repo, seems like only users logging in through OTP will be affected by this. 
If we cannot change the expiry time of google based logins too, I think keeping the same expiry time across all auth types is better, no?

deepansh96: No.
It works for google based logins too. 
The `ACCESS_TOKEN_EXPIRE_SECONDS` that I changed here is what is being used by both our custom OTP code and also by the rest_framework_social_oauth2 package.
I tested it out and it sets an expiry time of 7 days.

PR #335, opened 2022-02-25
Added check for survey questions in metrics calculation

### LINKED ISSUE #334: Adding survey mode to questions
**Describe the task**
Add `survey` column in `Question` table in db

Salonichoudhary07:
Fixes #334 

## Summary

survey questions will not be considered while calculating the metrics

## Test Plan

- [ ] Wrote tests
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production
    - [ ] update `docs/ENV.md`
- [x] If changes in DB, update [DB schema](https://docs.google.com/presentation/d/15c45HbHphh4w3guaFV3BgEA5nMKd-4J35JdUKdrpPzo/edit#slide=id.gc8c9ec0b5e_0_0) and BigQuery (staging and prod)


dalmia:
Screenshot 2022-02-28 at 9 49 45 AM
For the description in the DB schema, you need to write what "survey" means!

dalmia:
Screenshot 2022-02-28 at 9 50 24 AM
you have at least tested locally right?

plio/views.py
```diff
@@ -451,79 +451,94 @@ def metrics(self, request, uuid):
...
-                group_df = user_grouping.get_group(group)
-
-                # sanity check
-                assert num_questions == len(
-                    group_df
-                ), "Inconsistency in the number of questions"
-
-                num_answered = sum(
-                    group_df["answer"].apply(lambda value: value is not None)
+            dff = df[df["item_type"] == "question"].reset_index(drop=True)
```
dalmia, line 459: why create a separate data frame? you can simply update the `question_df` right?

plio/views.py
```diff
@@ -451,79 +451,94 @@ def metrics(self, request, uuid):
...
+            question_df = dff[dff.survey.eq(False)].reset_index(drop=True)
+            num_questions = len(question_df)
+
+            if num_questions == 0:
+                return Response(
+                    {
+                        "unique_viewers": num_unique_viewers,
+                        "average_watch_time": average_watch_time,
+                        "isQuesSurvey": True,
```
dalmia, line 468: there is an edge case here that will make this incorrect. think about it. Also, for different cases, we have returned all the metrics always at the end at once so that all the keys are present. We should keep that same so that the results always include all the metric keys.

dalmia, line 468: better name for `isQuesSurvey`

plio/views.py
```diff
@@ -451,79 +451,94 @@ def metrics(self, request, uuid):
...
-
-            user_grouping = question_df.groupby("user_id")
-            for group in user_grouping.groups:
-                # get the responses for a given user
-                group_df = user_grouping.get_group(group)
-
-                # sanity check
-                assert num_questions == len(
-                    group_df
-                ), "Inconsistency in the number of questions"
```
dalmia, line 480: why has this assertion been removed?

plio/views.py
```diff
@@ -451,79 +451,94 @@ def metrics(self, request, uuid):
...
+                        )
+
+                # converting to numpy arrays enabled us to use vectorization
+                # to speed up the computation many folds
+                num_answered_list = np.array(num_answered_list)
+                num_correct_list = np.array(num_correct_list)
+                average_num_answered = round(num_answered_list.mean())
+                percent_completed = np.round(
+                    100
+                    * (sum(num_answered_list == num_questions) / num_unique_viewers),
```
dalmia, line 509: should be on the same line

plio/views.py
```diff
@@ -451,79 +451,94 @@ def metrics(self, request, uuid):
...
+                    accuracy = None
+                else:
+                    accuracy = np.round(
+                        (num_correct_list / num_answered_list).mean() * 100, 2
+                    )
+
+                if num_questions != len(dff):
+                    quesSurvey = (True,)
+                else:
+                    quesSurvey = (False,)
```
dalmia, line 529: you can use a simpler and easier to read check: (question_df['survey'] == True).sum() > 0

plio/views.py
```diff
@@ -451,79 +451,94 @@ def metrics(self, request, uuid):
...
+
+                if not len(num_correct_list):
+                    accuracy = None
+                else:
+                    accuracy = np.round(
+                        (num_correct_list / num_answered_list).mean() * 100, 2
+                    )
+
+                if num_questions != len(dff):
+                    quesSurvey = (True,)
```
dalmia, line 527: why have you made it a tuple?

dalmia (changes requested):

dalmia:
Tests are failing

plio/views.py
Salonichoudhary07: because the questions will be inconsistent as survey mode is introduced and we do not include survey mode questions in metrics.

plio/views.py
Salonichoudhary07: The updated question_df does not have any `survey=True` so this will not work.

plio/views.py
Salonichoudhary07: I didn't, I tried so many times and researched for more than an hour why is it coming like that but could not find any solution.

plio/views.py
```diff
@@ -456,16 +456,22 @@ def metrics(self, request, uuid):
             )
 
             # retain only the responses to items which are questions
-            dff = df[df["item_type"] == "question"].reset_index(drop=True)
-            question_df = dff[dff.survey.eq(False)].reset_index(drop=True)
+            question_df = df[df["item_type"] == "question"].reset_index(drop=True)
+            question_df = question_df[question_df.survey.eq(False)].reset_index(
```
dalmia, line 460: comment saying that we are dropping survey questions

dalmia, line 460: it is more common syntax to simply do this and likely more computationally efficient too!
```suggestion
            question_df = question_df[~question_df.survey].reset_index(
```

plio/views.py
```diff
@@ -456,16 +456,22 @@ def metrics(self, request, uuid):
...
+                drop=True
+            )
             num_questions = len(question_df)
 
             if num_questions == 0:
```
dalmia, line 465:
```suggestion
            if not num_questions:
```

plio/views.py
```diff
@@ -456,16 +456,22 @@ def metrics(self, request, uuid):
...
                         "average_watch_time": average_watch_time,
-                        "isQuesSurvey": True,
+                        "percent_one_minute_retention": None,
+                        "accuracy": None,
+                        "average_num_answered": None,
+                        "percent_completed": None,
+                        "has_survey_question": True,
                     }
                 )
             else:
```
dalmia, line 477: once you have returned, you don't need to use else

plio/views.py
```diff
@@ -451,80 +451,100 @@ def metrics(self, request, uuid):
...
+
+                user_grouping = question_df.groupby("user_id")
+                for group in user_grouping.groups:
+                    # get the responses for a given user
+                    group_df = user_grouping.get_group(group)
+
+                    num_answered = sum(
+                        group_df["answer"].apply(lambda value: value is not None)
                     )
+                    # sanity check
```
dalmia, line 498:
```suggestion

                    # sanity check
```

dalmia (changes requested):

plio/views.py
```diff
@@ -451,80 +451,100 @@ def metrics(self, request, uuid):
                     "item_type",
                     "question_type",
                     "correct_answer",
+                    "survey",
```
dalmia, line 454: handle case when no responses have been given

plio/tests.py
```diff
@@ -943,16 +950,25 @@ def test_question_metrics_answers_provided(self):
...
+            options=["", ""],
+            correct_answer=0,
         )
         Question.objects.create(
             type="checkbox",
             item=item_2,
             text="test",
             options=["", ""],
             correct_answer=[0, 1],
+            survey=False,
```
deepansh96, line 965: by default survey mode is `False` right? so why write here explicitly? the tests would run fine even if you didn't add this here right?

plio/views.py
```diff
@@ -451,12 +458,30 @@ def metrics(self, request, uuid):
...
                 ],
             )
 
             # retain only the responses to items which are questions
             question_df = df[df["item_type"] == "question"].reset_index(drop=True)
-            num_questions = len(questions)
+            # retian only questions which are not survey questions
```
deepansh96, line 467:
```suggestion
            # retain only non-survey questions
```

plio/views.py
```diff
@@ -522,6 +547,10 @@ def is_answer_correct(row):
                 "accuracy": accuracy,
                 "average_num_answered": average_num_answered,
                 "percent_completed": percent_completed,
+                "has_survey_question": len(
+                    Question.objects.filter(item__plio=plio.id, survey=True)
```
deepansh96, line 551: reuse as a variable

plio/views.py
```diff
@@ -375,7 +375,14 @@ def metrics(self, request, uuid):
 
         # no sessions have been created for the plio: return
         if not Session.objects.filter(plio=plio.id):
-            return Response({})
+            return Response(
+                {
+                    "has_survey_question": len(
+                        Question.objects.filter(item__plio=plio.id, survey=True)
```
deepansh96, line 381: reuse as a variable

plio/views.py
```diff
@@ -451,12 +458,30 @@ def metrics(self, request, uuid):
...
+                    "survey",
                 ],
             )
 
             # retain only the responses to items which are questions
             question_df = df[df["item_type"] == "question"].reset_index(drop=True)
-            num_questions = len(questions)
+            # retian only questions which are not survey questions
+            question_df = question_df[~question_df.survey].reset_index(drop=True)
+            num_questions = len(
```
deepansh96, line 469:
```suggestion
            num_non_survey_questions = len(
```

plio/views.py
```diff
@@ -451,12 +458,30 @@ def metrics(self, request, uuid):
...
+                Question.objects.filter(item__plio=plio.id, survey=False)
+            )
+
+            if not len(question_df):
```
deepansh96, line 473: one line comment above this

plio/views.py
```diff
@@ -474,15 +499,15 @@ def is_answer_correct(row):
                 # get the responses for a given user
                 group_df = user_grouping.get_group(group)
 
+                num_answered = sum(
```
deepansh96, line 502: put this below the assertion check

deepansh96 (changes requested):

plio/views.py
```diff
@@ -373,9 +373,15 @@ def metrics(self, request, uuid):
         # else fetch the object
         plio = self.get_object()
 
+        num_survey_questions = Question.objects.filter(item__plio=plio.id, survey=True)
```
dalmia, line 376:
```suggestion
        survey_questions = Question.objects.filter(item__plio=plio.id, survey=True)
```

dalmia, line 376: instead of redefining `questions` later, do this here itself:
```
questions = Question.objects.filter(item__plio=plio.id)
survey_questions = questions.filter(survey=True)

File: plio/views.py [hunk 20]
```diff
@@ -451,12 +457,31 @@ def metrics(self, request, uuid):
...
                 ],
             )
 
             # retain only the responses to items which are questions
             question_df = df[df["item_type"] == "question"].reset_index(drop=True)
-            num_questions = len(questions)
+            # retain only non-survey questions
+            question_df = question_df[~question_df.survey].reset_index(drop=True)
+            num_non_survey_questions = len(
+                Question.objects.filter(item__plio=plio.id, survey=False)
```
dalmia, line 469: once you've defined `questions` as I have suggested above, just re-use it:
```suggestion
                questions.filter(survey=False)
```

plio/tests.py
```diff
@@ -777,12 +777,14 @@ def test_copying_to_workspace_with_no_video(self):
...
 
-    def test_metrics_returns_empty_if_no_sessions(self):
+    def test_no_metrics_returned_if_no_sessions(self):
         response = self.client.get(
             f"/api/v1/plios/{self.plio_1.uuid}/metrics/",
         )
+
         self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertEqual(response.data, {})
+        # if no sessions are there, no metrics will be returned only question mode will return
```
dalmia, line 786:
```suggestion
        # if no sessions are there, no metrics will be returned.
        # only whether any survey questions are there will be returned
```

plio/views.py
```diff
@@ -373,13 +373,16 @@ def metrics(self, request, uuid):
         # else fetch the object
         plio = self.get_object()
 
-        num_survey_questions = Question.objects.filter(item__plio=plio.id, survey=True)
+        # question-based metrics
```
dalmia, line 376: why has this comment been moved?

plio/views.py
```diff
@@ -465,9 +465,7 @@ def metrics(self, request, uuid):
             question_df = df[df["item_type"] == "question"].reset_index(drop=True)
             # retain only non-survey questions
             question_df = question_df[~question_df.survey].reset_index(drop=True)
-            num_non_survey_questions = len(
-                Question.objects.filter(item__plio=plio.id, survey=False)
-            )
+            non_survey_questions = len(questions.filter(survey=False))
```
dalmia, line 468: this is a number right?
```suggestion
            num_non_survey_questions = len(questions.filter(survey=False))
```

plio/views.py
Salonichoudhary07: Else there would be an error on line 379 ```survey_questions = questions.filter(survey=True)``` as the questions variable won't be defined

dalmia: how will a comment cause errors?

PR #338, opened 2022-04-07
Added the ETL app + bigquery_jobs table

### LINKED ISSUE #337: Add a table to keep a track of the changed data being synced from the DB to a data warehouse.
Refer to [this](https://www.notion.so/avantifellows/CDC-from-PostgreSQL-BQ-933c54d5b9664c11bfae9d0b58257210) doc to understand how the ETL pipeline will regularly sync data from this database and keep track of the data that has changed since the last sync process.

A special table is required to keep track of all the rows changed (inserted, updated, deleted) inside all the tables that need to be synced.

deepansh96:
Fixes #337 

## Summary

Refer to the issue linked above to read the doc regarding the ETL process that will regularly sync data from this database to a data warehouse.
This PR adds a table called `bigquery_jobs` to the database. The ETL process responsible for regularly syncing the data to a data warehouse will hook into this table for carrying out its operations.

## Test Plan

- [ ] ~Wrote tests~
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production
    - [ ] update `docs/ENV.md`
- [x] If changes in DB, update [DB schema](https://docs.google.com/presentation/d/15c45HbHphh4w3guaFV3BgEA5nMKd-4J35JdUKdrpPzo/edit#slide=id.gc8c9ec0b5e_0_0) and BigQuery (staging and prod)


plio/settings.py
```diff
@@ -86,6 +87,7 @@
     "rest_framework_social_oauth2",
     "storages",
     "silk",
+    "etl",
```
dalmia, line 90: why does it need to be in the tenant app given that we are storing the `schema` in the table inside the table in the `public` schema itself?

etl/serializers.py
```diff
@@ -0,0 +1,17 @@
...
+        model = BigqueryJobs
+        fields = [
+            "id",
+            "schema",
+            "table_to_sync",
+            "table_last_updated_at",
+            "table_last_synced_at",
+            "last_synced_row_id",
+            "created_at",
+            "updated_at",
```
dalmia, line 16: do we need these two in the serializer?

etl/views.py
```diff
@@ -0,0 +1,21 @@
...
+    retrieve: Retrieve a row from the bigquery jobs table
+    update: Update a row from the bigquery jobs table
+    create: Create a new row in bigquery jobs table
+    partial_update: Patch a row in bigquery jobs table
+    destroy: Soft delete a row of bigquery jobs table
+    """
+
+    queryset = BigqueryJobs.objects.all()
+    serializer_class = BigqueryJobsSerializer
+    permission_classes = [IsAuthenticated]
```
dalmia, line 21: shouldn't this be more tighter - like only superadmin?

etl/permissions.py
```diff
@@ -0,0 +1,15 @@
+from rest_framework import permissions
+
+
+class ETLPermissions(permissions.BasePermission):
```
dalmia, line 4: Check the docstring - has remnants of organization. Also, why should we keep has_permission as True and not keep it restricted to superuser as well?

etl/views.py
```diff
@@ -18,4 +19,4 @@ class BigqueryJobsViewSet(viewsets.ModelViewSet):
 
     queryset = BigqueryJobs.objects.all()
     serializer_class = BigqueryJobsSerializer
-    permission_classes = [IsAuthenticated]
+    permission_classes = [IsAuthenticated, ETLPermissions]
```
dalmia, line 22: IsAuthenticated is redundant now no?

PR #343, opened 2023-10-05
(4) added a new table to csv data - user-level-metrics.csv

### LINKED ISSUE #344: Requests by Nudge -- Add more info to dowloaded csv data
- We're removing `unique_id:` prefix from the `user_identifier` column values. Instead of this, a new column called `has_user_logged_in_via_sso` is added which will tell whether this user came via sso or the default email/phone login.
- adding `created_at` and `updated_at` timestamps to the sessions.csv
- adding created at time as `answered_at` to the entries of response.csv, signifying at what time an answer was recorded.
- adding a new table `user-level-metrics.csv`. This will contain all the users, how many questions did they attempt, did they attempt all the questions, number of correctly answered questions.
- Adding `correct_answer` and `is_answer_correct` to the `responses.csv`
- update the README

deepansh96:
Fixes #344


plio/queries.py
```diff
@@ -190,3 +201,79 @@ def get_events_query(plio_uuid: str, schema: str, show_unmasked_user_id: bool =
...
+    """
+    Returns data for each user for the given plio. These include
+    - The user's identifier
+    - Whether the user has logged in via SSO
+    - Number of questions attempted by the user
+    - Number of questions answered correctly by the user
+    - Number of questions answered incorrectly by the user
+    - Number of questions skipped by the user
+    - Has the user attempted all the questions?
+    - Has the user answered all the questions correctly?
```
pritamps, line 218: Why is this needed. Looks like it's already there from the other columns?

deepansh96: So we were not giving them the correct answer before. If they had correct answer, they can derive these values. But that would require them to take it into excel or pandas and do some joins.

Usually non tech people use these csvs from the plio dashboard, so that's why this info might make sense to be written into the file itself.

Also nudge people specifically asked for this :p

PR #349, opened 2025-04-18
fix #47: after pip install update requirement.txt

### LINKED ISSUE #47: Auto detect new dependencies and update requirements.txt
**Describe the task**
Adding new dependencies is a manual step right now. The developer needs to run the following command in order to stay up to date:
```
pip freeze > requirements.txt
```

**Expected behavior**
We need an automated way that automatically captures when a new package is installed (maybe when we run `pip install`) and adds it to requirements.txt. Something like npm does when we use npm install, it adds to `package.json` and `package-lock.json`

amit-dhidhi-dev:
step 1: setup the project in local machine.
step 2: check it. project setup is correct by running it.
step 3:  create scripts folder and add this three files. 
          1. pip.ps1
          2. pip.bat
          3. pip-install.sh
which created in this PR. this file capable of update requirements.txt when packages are installed successfully.
for this in the **terminal**  set this .
        1.  for windows - set alias like : Set-Alias -Name pip -Value "$PWD\scripts\pip.ps1"
        or 
         1.  for linux - set alias like : alias pip="$PWD/scripts/pip-install.sh"
         
after this alias set up 
  when you install packages through " pip install <package> " it will add the " package name== package version" in the requirements.txt file. 

if you not set the alias - you should run like "scripts\pip install package_name"(windows) and for linux "scripts\pip-install install package_name " for installing packages. then this will update the requirements.txt

### DIFF
--- a/plio/settings.py
+++ b/plio/settings.py
@@ -14,6 +14,10 @@ import os
 import logging
 from corsheaders.defaults import default_headers
 
+# Load environment variables from .env file
+from dotenv import load_dotenv
+load_dotenv()
+
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,33 @@
+# Django Project Package Management
+
+This directory contains scripts to help manage Python packages in this Django project.
+
+## Usage
+
+### Windows (PowerShell or Command Prompt)
+
+Use `scripts\pip.bat` instead of `pip`:
+
+```
+scripts\pip install package_name
+```
+
+This will:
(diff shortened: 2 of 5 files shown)


amit-dhidhi-dev:
step1:
for win: set alias like :  Set-Alias -Name pip -Value "$PWD\scripts\pip.ps1"
for linux:  set alias like :  alias pip="$PWD/scripts/pip-install.sh"
step 2: after that when you install pip install <package> and installed successfully it will update requirement.txt

deepansh96:
> step1: for win: set alias like : Set-Alias -Name pip -Value "$PWD\scripts\pip.ps1" for linux: set alias like : alias pip="$PWD/scripts/pip-install.sh" step 2: after that when you install pip install and installed successfully it will update requirement.txt

Please write clear instructions on what this PR is about by editing the PR descriptions. These instructions are not super clear

Issue #443, opened 2026-09-17
Staging real-browser end-to-end acceptance: 56-scenario plan and preflight checklist


deepansh96:
## Objective

Determine whether the **deployed Plio staging environment** works end to end through real browser interaction, across creator, learner and workspace-administration flows. Exercise real login, YouTube playback, uploads, downloads and saved state; existing API/ephemeral-CI tests alone do not establish staging readiness.

Scope spans **plio-frontend and plio-backend**. This backend issue is the shared tracker, following test-wall map #362 and workstreams #374–#380. It contains the complete **56-scenario plan: 30 critical and 26 extended**, plus preflight requirements and explicit evidence/pass criteria. This issue is a plan; it does not authorize a production release.

## Execution tracking

- [ ] Complete and publish P01–P10 preflight outcomes below.
- [ ] Prepare the staging-only browser runner and isolated fixture ledger.
- [ ] Execute all 30 critical scenario groups, recording each browser/role variant.
- [ ] Execute all 26 applicable extended scenario groups and separately report privileged/backend-only supplements.
- [ ] File/link defects with redacted evidence; rerun affected cases and the core journey after fixes.
- [ ] Complete run-owned fixture cleanup and publish the final verdict, with every blocked/not-run case visible.

## Preflight: complete before the acceptance run

Preflight establishes the target, access, tools, isolated data and evidence collection. It does **not** mean the product passed. All checks below start unchecked; September 17 reconnaissance is a dated observation, not a substitute for rechecking at execution time.

For each item, record `PASS / FAIL / BLOCKED`, date, evidence and an owner for anything unresolved. Run read-only checks first. Provision only dedicated test fixtures after confirming the target. Do not redeploy, modify infrastructure, globally clear caches, reset shared users or wipe OAuth tokens as an implicit preflight repair.

### P01 — Identify the deployment being tested

- [ ] Confirm frontend `https://staging-app.plio.in/` and backend `https://staging-backend.plio.in/` are the intended targets; verify actual browser API/auth calls use staging rather than production or localhost.
- [ ] Record the running frontend and backend versions using serving asset/build identifiers and ECS task image digests/task definitions, mapped to commit SHAs where possible. Record current deployment/health state through read-only tooling. A workflow's success/failure or the current `main` SHA alone is insufficient.
[...]

PR #444, opened 2026-09-17
Fix DEBUG environment parsing

deepansh96:
## Problem

`DEBUG` was assigned directly from `os.environ.get`, so any nonempty string such as `DEBUG=False` or `DEBUG=0` evaluated as truthy and enabled Django debug pages. This was observed in staging and production during the P02 preflight tracked by [#443](https://github.com/avantifellows/plio-backend/issues/443).

## Change

- Parse `DEBUG` as a boolean: `True`/`1` enable debug mode and `False`/`0` disable it, ignoring case and surrounding whitespace.
- Treat unset, empty, and unrecognized values as `False` so a malformed deployment value cannot enable debug mode.
- Document the accepted values and fallback behavior in `docs/ENV.md`.
- Add regression coverage that imports the actual settings module in a fresh process for unset, false, zero, true, one, whitespace/case variants, and invalid values.

## Validation

- `pre-commit run --all-files --show-diff-on-failure`
- `/tmp/plio-debug-fix/venv312/bin/pytest --confcutdir=tests tests/test_settings.py -q` (14 passed)
- `python -m py_compile plio/settings.py tests/test_settings.py`
- `git diff --check`

The full Django unit and integration lanes were not run locally because the Docker daemon was unavailable; CI should run the repository's supported database and Redis-backed suites.

## Impact

This changes only environment parsing for Django's `DEBUG` setting. Existing explicit true values remain enabled, while the observed string values `False` and `0` now correctly disable debug output. No staging or production environment was changed.

Related to #443 (P02 preflight context).


deepansh96:
@codex review

PR #445, opened 2026-09-17
Serve collected static files through ASGI

deepansh96:
## Problem

PR [#444](https://github.com/avantifellows/plio-backend/pull/444) correctly makes `DEBUG=False` disable Django debug mode. In that production mode, Daphne still sends HTTP requests directly to Django's ASGI handler, while `collectstatic` writes the docs and admin assets to `/app/static`; Django's development static helper is disabled, so the ReDoc document loads without its JavaScript/CSS and the admin login is unstyled.

This is a stacked follow-up to #444. Merge #444 first; #444 itself is unchanged. The staging failure and release context are tracked in [#443](https://github.com/avantifellows/plio-backend/issues/443).

## Change

- Pin `servestatic==4.3.4` and wrap only the existing HTTP Django ASGI branch with `ServeStaticASGI`, rooted at `settings.STATIC_ROOT` and mounted at `settings.STATIC_URL`.
- Keep the Channels WebSocket branch, early Django setup, existing storage backends, and startup collection unchanged in shape. `collectstatic --no-input --clear` runs before Daphne against the container-local generated tree.
- Set a 60-second cache lifetime for unversioned assets and add `X-Content-Type-Options: nosniff` through ServeStatic's documented header callback. User uploads remain on their configured storage backend.
- Ignore only the repository-root generated `/static/` directory in Docker build context, leaving `plio/static/` source assets included.
- Document the deployment assumption and add a regression test against the actual `plio.asgi.application` with `DEBUG=False`, temporary collected assets, byte/MIME/header/HEAD assertions, missing/traversal/repository-root checks, and API fall-through.

## Validation

- `pre-commit run --all-files --show-diff-on-failure`
- `pytest --ignore=tests/integration -q`: 321 passed, 1 expected failure
- `pytest -m integration -q`: 131 passed, 1 skipped
- Coverage floor checks: unit 93.1% vs 92.15%; integration 80.5% vs 79.35%
- Real local Daphne smoke after `collectstatic --no-input --clear`: `insQ.min.js`, `redoc-init.js`, `redoc/redoc.min.js`, `style.css`, and `admin/css/base.css` all returned byte-for-byte matching `200` responses with correct MIME, `nosniff`, cache, and validator headers; HEAD, missing, encoded traversal, NUL, API 401, and repository-root probes were safe.
- Playwright/Chromium smoke: ReDoc rendered visible `Plio API` content with all four docs assets at `200`; the admin login rendered with `admin/css/base.css` at `200` and expected admin typography.

The local checks used isolated pytest PostgreSQL/Redis resources. Docker Desktop was unavailable, so the built-image/Docker Compose path and staging/production browser smoke remain CI/release-gate work. No staging or production state was changed.

## Rollout

After #444 and this follow-up receive fresh review, deploy the exact combined head to staging and repeat the direct asset, missing/traversal, docs-rendering, admin-rendering, API, and WebSocket checks before production approval.


deepansh96:
@codex review

